PR #2, opened 2020-05-14
(DOCSP-10354): Landing page

jeff-allen-mongo:
Hey @mmarcon, I'd appreciate your eyes on this first draft of the landing page for the MongoDB Shell.

Some notes for context:

- This content is largely modeled after the [existing shell docs](https://docs.mongodb.com/manual/mongo/)

- We will be handling the download / installation instructions in a separate ticket, as well as fleshing out the "Learn More" section at a later date.

- Included in this draft is a skeleton for how we want to shape the rest of the docs. Even though we are linking out to the existing manual docs for command reference, I think it would be helpful to include a baseline "Getting Started" with basic CRUD / aggregation.

Some questions:

1. Just confirming the product name is "MongoDB Shell"? Asking because in the [product wiki](https://github.com/mongodb-js/mongosh/wiki) it is `mongosh`. If we want to call it `mongosh`, it's an easy fix.

2. Is there anything else we should call out / highlight on this page?

3. Looking for specific feedback on the [Comparison section](https://docs-mongodbcom-staging.corp.mongodb.com/ac7ecc3/mongodb-shell/docsworker/DOCSP-10354/#comparison-of-the-mdb-shell-and-legacy-mongo-shell), and if there is anything else we want to say there.

**Staged**: [MongoDB Shell](https://docs-mongodbcom-staging.corp.mongodb.com/266ebf3/mongodb-shell/docsworker/DOCSP-10354/)

### DIFF
--- a/source/index.txt
+++ b/source/index.txt
@@ -45,8 +45,8 @@ Download and Install the |mdb-shell|
 Multi-Line Operations in the |mdb-shell|
 ----------------------------------------
 
-Run the ``editor()`` command inside of the |mdb-shell| to open the
-editor and manage comprehensive multiline functions:
+Run the ``.editor`` command within the |mdb-shell| to open the editor
+and manage comprehensive multiline functions:
 
 .. example::
 
@@ -54,11 +54,12 @@ editor and manage comprehensive multiline functions:
 
    .. code-block:: javascript
 
-      editor()
+      .editor
 
    The |mdb-shell| enters editor mode:
 
    .. code-block:: javascript
+      :copyable: false
 
       // Entering editor mode (^D to finish, ^C to cancel)
       const { Writable } = require('stream');
@@ -72,10 +73,18 @@ editor and manage comprehensive multiline functions:
           }
         });
 
+- Press ``Cmd/Ctrl`` + ``d`` to finish and run your function.
+
+- Press ``Cmd/Ctrl`` + ``c`` to cancel and exit without running your
+  function.
+
+When you close the editor, the |mdb-shell| does not save any functions
+you had entered into the editor.
+
 .. note::
 
-   Currently, ``editor()`` is the only way to run multiline commands in
-   the |mdb-shell|.
+   Currently, ``editor()`` is the only way to paste multiline commands
+   into the |mdb-shell|.
 
 
 Comparison of the |mdb-shell| and Legacy ``mongo`` Shell
--- a/source/index.txt
+++ b/source/index.txt
@@ -14,10 +14,9 @@ MongoDB Shell
 
 .. include:: /includes/admonitions/fact-mdb-shell-beta.rst
 
-The |mdb-shell| is an interactive JavaScript shell interface to MongoDB,
-which provides a powerful interface for users to test queries and
-operations directly with the database. The |mdb-shell| also provides a
-fully functional JavaScript environment for use with MongoDB.
+The |mdb-shell| is a fully functional JavaScript environment for
+interacting with MongoDB deployments. You can use the |mdb-shell| to
+test queries and operations directly with your database.
 
 The |mdb-shell| is available as a standalone package in the MongoDB
 download center.
@@ -87,8 +86,8 @@ you had entered into the editor.
    into the |mdb-shell|.
 
 
-Comparison of the |mdb-shell| and Legacy ``mongo`` Shell
---------------------------------------------------------
+The |mdb-shell| versus the Legacy ``mongo`` Shell
+-------------------------------------------------
 
 The |mdb-shell| offers numerous advantages over the legacy
 :binary:`mongo <mongo>` shell, such as:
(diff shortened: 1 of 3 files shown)


source/index.txt
```diff
@@ -10,6 +12,96 @@ MongoDB Shell
...
+   .. code-block:: javascript
+
+      editor()
+
+   The |mdb-shell| enters editor mode:
+
+   .. code-block:: javascript
+
+      // Entering editor mode (^D to finish, ^C to cancel)
+      const { Writable } = require('stream');
```
jeff-allen-mongo, line 64: [note] need to make this `:copyable: false`

source/index.txt
```diff
@@ -10,6 +12,96 @@ MongoDB Shell
...
+
+   .. tab::
+      :tabid: linux
+      
+      Hello, I am Linux (rpm / deb) content.
+
+Multi-Line Operations in the |mdb-shell|
+----------------------------------------
+
+Run the ``editor()`` command inside of the |mdb-shell| to open the
```
mmarcon, line 48:
```suggestion
Run the ``.editor`` command inside of the |mdb-shell| to open the
```

source/index.txt
```diff
@@ -10,6 +12,96 @@ MongoDB Shell
...
+editor and manage comprehensive multiline functions:
+
+.. example::
+
+   Run the following command within the |mdb-shell|:
+
+   .. code-block:: javascript
+
+      editor()
```
mmarcon, line 57:
```suggestion
      .editor
```

source/index.txt
```diff
@@ -10,6 +12,96 @@ MongoDB Shell
...
+            // ...
+          },
+          writev(chunks, callback) {
+            // ...
+          }
+        });
+
+.. note::
+
+   Currently, ``editor()`` is the only way to run multiline commands in
```
mmarcon, line 77:
```suggestion
   Currently, ``editor()`` is the only way to paste multiline commands in
```

mmarcon: You can do multiline editing also without `.editor`, it's pasting a multiline command/script that does not work.

mmarcon:
Fixed the product name in the product wiki. Thank you!

jeff-allen-mongo:
Thanks for the feedback @mmarcon. I fixed those commands

source/index.txt
```diff
@@ -10,6 +12,105 @@ MongoDB Shell
...
+The |mdb-shell| offers numerous advantages over the legacy
+:binary:`mongo <mongo>` shell, such as:
+
+- Improved syntax highlighting.
+
+- Improved command history.
+
+- Improved logging.
+
+During the beta phase, the |mdb-shell| supports a subset of the legacy
```
zach-carr, line 102: [nit] In the note it's called "Beta stage," suggest going with one or the other

source/index.txt
```diff
@@ -10,6 +12,105 @@ MongoDB Shell
    :depth: 1
    :class: singlecol
 
+.. include:: /includes/admonitions/fact-mdb-shell-beta.rst
+
+The |mdb-shell| is an interactive JavaScript shell interface to MongoDB,
+which provides a powerful interface for users to test queries and
+operations directly with the database. The |mdb-shell| also provides a
+fully functional JavaScript environment for use with MongoDB.
```
zach-carr, line 20: [suggestion] I think this could be simplified some, maybe something like:

The |mdb-shell| is an interactive JavaScript environment for interfacing with a MongoDB deployment. You can use the shell to test queries and other operations directly with a database.

source/index.txt
```diff
@@ -10,6 +12,105 @@ MongoDB Shell
...
+When you close the editor, the |mdb-shell| does not save any functions
+you had entered into the editor.
+
+.. note::
+
+   Currently, ``editor()`` is the only way to paste multiline commands
+   into the |mdb-shell|.
+
+
+Comparison of the |mdb-shell| and Legacy ``mongo`` Shell
```
zach-carr, line 90: [suggestion] "|mdb-shell| vs. Legacy ``mongo`` Shell" is more immediately clear to me as a section topic. Up to you, I don't know if the style guide says anything about using/not using "vs."

source/index.txt
```diff
@@ -10,6 +12,105 @@ MongoDB Shell
...
 
+.. include:: /includes/admonitions/fact-mdb-shell-beta.rst
+
+The |mdb-shell| is an interactive JavaScript shell interface to MongoDB,
+which provides a powerful interface for users to test queries and
+operations directly with the database. The |mdb-shell| also provides a
+fully functional JavaScript environment for use with MongoDB.
+
+The |mdb-shell| is available as a standalone package in the MongoDB
+download center.
```
zach-carr, line 23: Link to download center

source/index.txt
jeff-allen-mongo: Agree. This was largely borrowed from the existing copy in the manual, but it could be simplified.

source/index.txt
jeff-allen-mongo: Tweaked this title.

source/index.txt
jeff-allen-mongo: Will do, once it is available there.

PR #3, opened 2020-05-18
(DOCSP-10363): Add available methods

jeff-allen-mongo:
Staged: [Available Methods](https://docs-mongodbcom-staging.corp.mongodb.com/eb3d0c6/mongodb-shell/docsworker/DOCSP-10363/reference/methods)

Hey @mmarcon, in looking at the [output of the available new shell methods](https://mongodb.slack.com/files/UD96QJACS/F014EKV00BA/untitled?origin_team=T024FNNHU&origin_channel=Vall_threads) there are some methods in the output which we do not document on the [legacy mongo shell docs](https://docs.mongodb.com/manual/reference/method/).

I wanted to list them here just to see if they were maybe new methods, or wrappers for some other method?

* Collection Methods
  * convertToCapped (this looks to be a [command](https://docs.mongodb.com/manual/reference/command/convertToCapped/), so I'm wondering if there is now a wrapper for this command?)
  * exists
  * getIndexKeys
  * getIndexSpecs

* Cursor Methods
  * arrayAccess
  * clone
  * oplogReplay
  * projection

### DIFF
--- a/source/reference.txt
+++ b/source/reference.txt
@@ -10,5 +10,5 @@ Reference
       :titlesonly:
 
       /reference/settings
-      /reference/commands
+      /reference/methods
       /reference/logging
--- a/source/reference/commands.txt
+++ b/source/reference/commands.txt
@@ -1,9 +1,0 @@
-=======================================
-Available Commands in the MongoDB Shell
-=======================================
-
-.. contents:: On this page
-   :local:
-   :backlinks: none
-   :depth: 1
-   :class: singlecol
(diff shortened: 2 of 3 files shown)


mmarcon:
They are not new. I just checked them all and they all exist in `mongo`.

`convertToCapped` does indeed just invoke the corresponding command.

```
> db.xyz.convertToCapped
function(bytes) {
    if (!bytes)
        throw Error("have to specify # of bytes");
    return this._dbCommand({convertToCapped: this._shortName, size: bytes});
}
```

I am ok with leaving them undocumented, but I wanted if leaving them undocumented for the old shell was intentional.

PR #7, opened 2020-05-21
(DOCSP-10356) Windows install procedure

zach-carr:
[Ticket](https://jira.mongodb.org/browse/DOCSP-10356)
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/bb5b8b4/mongodb-shell/docsworker/DOCSP-10356/#download-and-install-the-mdb-shell)

Question for review: should there be an unzip procedure in step 4? I decided not to include one because unzipping can be done in many simple ways, including dragging the contents of the .zip folder out of the folder. Covering multiple Windows unzipping methods seems outside of scope and choosing one that works for all Windows versions overcomplicates later versions.

The procedure in step 5 covers Windows 10, 8, and 7. It's not very different for Vista and XP.

### DIFF
--- a/source/includes/steps-install-shell-windows.yaml
+++ b/source/includes/steps-install-shell-windows.yaml
@@ -18,24 +18,8 @@ title: "Extract the files from the downloaded archive."
 ref: extract-archive
 level: 4
 content: |
-   
-   To extract files from a ``.zip`` folder:
-
-   1. Right click on the folder you want to extract files from.
-
-   #. Select :guilabel:`Extract All...`
-
-   #. Choose a file destination to which the folder's contents are 
-      extracted. The default destination is the same directory as the 
-      ``.zip`` folder.
-
-   #. Click :guilabel:`Extract`.
-
-   The contents of the ``.zip`` folder are extracted to the specified 
-   destination.
-
-   Skip this step if your web browser automatically unzips the file as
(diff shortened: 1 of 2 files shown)


source/includes/steps-install-shell-windows.yaml
```diff
@@ -0,0 +1,60 @@
...
+      :guilabel:`Edit`. The **Edit environment variable** modal 
+      displays.
+
+   #. Click :guilabel:`New` and add the filepath to your ``mongosh`` 
+      binary.
+
+   #. Click :guilabel:`OK` to confirm your changes. On each other 
+      modal, click :guilabel:`OK` to confirm your changes.
+
+   To confirm that your ``PATH`` environment variable is correctly
```
MongoCaleb, line 54: Since you don't specify how to extract the archive in step 4, I'd also consider omitting how to open a command prompt, since there are multiple ways to do that, too. But perhaps I'm just used to doing it and it's not as simple as I think?

zach-carr: Good call, I agree. Removed that part.

zach-carr:
Thanks Caleb - I forgot to mention that this repo is still in the works and that steps 1-2 are subject to change and the download isn't actually available there yet.

PR #8, opened 2020-05-26
(DOCSP-10361): Add mongodb shell options

jeff-allen-mongo:
Hey @mmarcon, this doesn't need a full review since these descriptions are copied from the existing manual descriptions, however there were a couple options listed in the [readme](https://github.com/mongodb-js/mongosh#cli-usage) which were not documented on the manual side, so I wanted to see what you thought.

If you can provide a brief description for these I'm happy to add them if you'd like:

- --disableImplicitSessions

- --kmsURL

One other note, our next-gen docs platform currently does not support cross-linking options together (e.g., when we say "where the specified `--username` has been created", linking `--username` to the `--username` option) but this is something I'm discussing with our platform team to see if we can make happen.

Staging: [MongoDB Shell Options](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker/DOCSP-10361-v3/reference/options)


mmarcon:
Apologies for the very late review!
Those are FLE and Transaction options, so they are not needed in this first release anyways.

jeff-allen-mongo:
@mmarcon Sounds good. I'll merge these updates in that case.

PR #9, opened 2020-05-27
(DOCSP-10504): CRUD ops for mongosh

jeff-allen-mongo:
Staged: [Perform CRUD Operations](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker/DOCSP-10504/crud) plus sub-pages.

Questions for review

- What do you think about the flow here where we highlight the Atlas sample data? I wanted to insert data into a fresh collection, so they at least have something to work with if they don't want to spin up an Atlas cluster.

- There is a little bit of inconsistency between using `example` blocks and going with "For example..." prose. Lmk if you have a preference or if you think we should standardize more than what's done here.

### DIFF
--- a/source/crud/insert.txt
+++ b/source/crud/insert.txt
@@ -70,10 +70,10 @@ adds the ``_id`` field with an ObjectId value to each document. See
    ])
 
 :method:`~db.collection.insertMany()` returns a document that includes
-the newly inserted documents ``_id`` field values.
+the newly inserted documents' ``_id`` field values.
 
 To retrieve the inserted documents,
-:ref:`read the collection <mongosh-read>`:
+:ref:`read documents in the collection <mongosh-read>`:
 
 .. code-block:: javascript
  
--- a/source/crud/insert.txt
+++ b/source/crud/insert.txt
@@ -22,32 +22,52 @@ into a collection:
 - To insert multiple documents, use
   :method:`db.collection.insertMany()`.
 
+.. include:: /includes/fact-sample-data-examples.rst
+
 Insert a Single Document
 ------------------------
 
 :method:`db.collection.insertOne()` inserts a *single*
 :ref:`document<bson-document-format>` into a collection.
 
-The following example inserts a new document into the ``inventory``
-collection. If the document does not specify an ``_id`` field, MongoDB
-adds the ``_id`` field with an ObjectId value to the new document. See
-:ref:`write-op-insert-behavior`.
+.. example::
 
-.. code-block:: javascript
+  This example inserts a new document into the ``sample_mflix.movies``
+  collection. If the document does not specify an ``_id`` field, MongoDB
+  adds the ``_id`` field with an ObjectId value to the new document. See
+  :ref:`write-op-insert-behavior`.
 
-   db.inventory.insertOne(
-      { item: "canvas", qty: 100, tags: ["cotton"], size: { h: 28, w: 35.5, uom: "cm" } }
-   )
+  .. code-block:: javascript
 
-:method:`~db.collection.insertOne()` returns a document that
-includes the newly inserted document's ``_id`` field value.
+     use sample_mflix
 
-To retrieve the inserted document,
-:ref:`read the collection <mongosh-read>`:
+     db.movies.insertOne(
+       {
+         title: "The Favourite",
+         genres: [ "Drama", History" ]
+         runtime: 121,
+         rated: "R",
+         year: 2018,
+         directors: [ "Yorgos Lanthimos" ],
+         cast: [ "Olivia Colman", "Emma Stone", "Rachel Weisz" ],
+         type: "movie"
+       }
+     )
 
-.. code-block:: javascript
+   :method:`~db.collection.insertOne()` returns a document that
+   includes the newly inserted document's ``_id`` field value.
 
-   db.inventory.find( { item: "canvas" } )
+   To retrieve the inserted document,
+   :ref:`read the collection <mongosh-read>`:
+
+   .. code-block:: javascript
+
+      db.inventory.find( { title: "The Favourite" } )
+
+   .. note::
+
+      To ensure you return the document you inserted, you can instead
+      query by ``_id``.
 
 Insert Multiple Documents
 -------------------------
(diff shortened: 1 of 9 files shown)


source/crud/update.txt
```diff
@@ -0,0 +1,153 @@
...
+   db.accounts.replaceOne(
+     { account_id: 371138 },
+     { account_id: 893421, limit: 5000, products: [ "Investment", "Brokerage" ] }
+   )
+
+Run the following command to read your updated document:
+
+.. code-block:: javascript
+
+   db.accounts.findOne( { account_id: 371138 } )
```
jeff-allen-mongo, line 137: note for review: this won't work because we updated the account id. Will fix in patch 2.

source/crud/read.txt
```diff
@@ -0,0 +1,166 @@
...
+
+Use the :method:`db.collection.find()` method in the |mdb-shell|
+to query documents in a collection.
+
+.. include:: /includes/fact-sample-data-examples.rst
+
+.. TODO: Make a callout here to the Connect to an Atlas Cluster section
+   in the connection docs
+
+Select All Documents in a Collection
```
jwilliams-mongo, line 25: Is this section titled this way because of the relationship to the SQL operation? Seems like we use `query` elsewhere.

source/crud/read.txt
```diff
@@ -0,0 +1,166 @@
...
+------------------------------------
+
+To select all documents in the collection, pass an empty document as the
+query filter parameter to the find method. The query filter parameter
+determines the select criteria.
+
```
jwilliams-mongo, line 31: This seems to return only 20 documents if the collection has more than 20. Behavior seems different from the legacy shell where you can type `it` to view additional documents. Might be worth noting.

source/crud/read.txt
```diff
@@ -0,0 +1,166 @@
...
+      use sample_mflix
+
+      db.movies.find( { "title": "Titanic" } )
+
+   This operation corresponds to the following SQL statement:
+
+   .. code-block:: sql
+      :copyable: false
+
+      SELECT * FROM movies WHERE title = "Titanic"
```
jwilliams-mongo, line 72: Are these pulled from the manual? Do we have corresponding SQL commands for the commands in the other tutorials?

source/crud/read.txt
```diff
@@ -0,0 +1,166 @@
...
+   .. code-block:: javascript
+
+      use sample_mflix
+
+      db.movies.find( {
+           year: 2010,
+           $or: [ { "awards.wins": { $gte: 5 } }, { genres: "Drama" } ]
+      } )
+
+Additional Query Tutorials
```
jwilliams-mongo, line 149: Should we also have a `Behavior` section here like for the other tutorials?

source/crud/update.txt
```diff
@@ -0,0 +1,153 @@
...
+   :depth: 1
+   :class: singlecol
+
+.. include:: /includes/admonitions/fact-mdb-shell-beta.rst
+
+The MongoDB shell provides the following methods to update documents in
+a collection:
+
+- To update a single document, use
+  :method:`db.collection.updateOne(\<filter\>, \<update\>, \<options\>)
```
jwilliams-mongo, line 21: we don't include the parameters for the methods in the other tutorials. not sure we should include them here to be consistent.

source/crud/update.txt
```diff
@@ -0,0 +1,153 @@
...
+details.
+
+Update a Single Document
+------------------------
+
+The following example uses the :method:`db.collection.updateOne()`
+method on the ``sample_mflix.movies`` collection to update the *first*
+document where ``title`` equals ``"Titanic"``:
+
+.. code-block::
```
jwilliams-mongo, line 64: see comments above about example blocks.

source/crud/update.txt
```diff
@@ -0,0 +1,153 @@
...
+   db.movies.updateOne( { title: "Titanic" },
+   {
+     $set: { countries: [ "USA", "Canada" ] },
+     $currentDate: { lastUpdated: true }
+   }
+
+The update operation:
+
+- uses the :update:`$set` operator to update the value of the
+  ``countries`` array to ``[ "USA", "Canada" ]``,
```
jwilliams-mongo, line 77:
```suggestion
- Uses the :update:`$set` operator to update the value of the
  ``countries`` array to ``[ "USA", "Canada" ]``.
```

source/crud/update.txt
```diff
@@ -0,0 +1,153 @@
...
+
+- uses the :update:`$currentDate` operator to update the value
```
jwilliams-mongo, line 79:
```suggestion
- Uses the :update:`$currentDate` operator to update the value
```

source/crud/read.txt
jeff-allen-mongo: Whoops missed this one. Updating.

source/crud/insert.txt
```diff
@@ -0,0 +1,97 @@
...
+:method:`db.collection.insertMany()` can insert *multiple*
+:ref:`documents <bson-document-format>` into a collection. Pass an array
+of documents to the method.
+
+The following example inserts three new documents into the ``inventory``
+collection. If the documents do not specify an ``_id`` field, MongoDB
+adds the ``_id`` field with an ObjectId value to each document. See
+:ref:`write-op-insert-behavior`.
+
+.. code-block:: javascript
```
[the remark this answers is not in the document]
jwilliams-mongo: @jeff-allen-mongo missed an example block here

source/crud/insert.txt
```diff
@@ -0,0 +1,133 @@
...
+         type: "movie"
+       }
+   ])
+
+:method:`~db.collection.insertMany()` returns a document that includes
+the newly inserted documents' ``_id`` field values.
+
+To :ref:`read documents in the collection <mongosh-read>`:
+
+.. code-block:: javascript
```
jwilliams-mongo, line 114: example block?

source/crud/read.txt
jwilliams-mongo: don't think it's crucial to have these now.

source/crud/read.txt
jeff-allen-mongo: cc @mmarcon, this PR is already merged but I forgot to ask about this. When a read operation returns more than 20 docs in the mongodb shell, are we expected to be able to iterate over those docs with a cursor? if not, what is the expected behavior?

PR #10, opened 2020-05-29
(DOCSP-10355) Connect to a Deployment

zach-carr:
[Ticket](https://jira.mongodb.org/browse/DOCSP-10355)
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker/DOCSP-10355/connect)

This page will be updated with --host and --port options when https://jira.mongodb.org/browse/MONGOSH-219 is resolved.

### DIFF
--- a/source/connect.txt
+++ b/source/connect.txt
@@ -15,8 +15,8 @@ Connect to a Deployment
 
 .. important::
 
-   Ensure that a MongoDB deployment is running before attempting to 
-   connect to it with ``mongosh``. The connection method you use 
+   Ensure that you have a running MongoDB deployment before attempting 
+   to connect with ``mongosh``. The connection method you use 
    depends on your deployment type.
 
 Local MongoDB Instance on Default Port
@@ -29,6 +29,12 @@ to connect to a MongoDB instance running on your **localhost** with
 .. code-block:: sh
 
    mongosh
+
+This is equivalent to the following command:
+
+.. code-block:: sh
+
+   mongosh "localhost:27017"
 
 Local MongoDB Instance on a Non-Default Port
 --------------------------------------------
@@ -66,11 +72,11 @@ To specify a remote host and port, use a
    your connection string from the |service| UI. To learn more, see 
    :atlas:`Connect to a Cluster <connect-to-cluster/#use-the-connect-dialog-to-connect-to-your-cluster>`.
 
-Connection Considerations
--------------------------
+Connection Options
+------------------
 
-Authentication
-~~~~~~~~~~~~~~
+Connect with Authentication
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 To connect to a MongoDB instance requires authentication, use the 
 ``--username`` and ``--authenticationDatabase`` command-line options. 
@@ -97,8 +103,8 @@ To connect to a MongoDB instance requires authentication, use the
    - Add :manual:`Database Users </core/security-users/>` 
      to provide authenticated access to a MongoDB instance.
 
-Connecting to a Replica Set
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Connect to a Replica Set
+~~~~~~~~~~~~~~~~~~~~~~~~
 
 To connect to a replica set:
 
@@ -130,23 +136,10 @@ To connect to a replica set:
 
         mongosh "mongodb://mongodb0.example.com.local:27017,mongodb1.example.com.local:27017,mongodb2.example.com.local:27017/?replicaSet=replA"
 
-TLS Connections
-~~~~~~~~~~~~~~~
+Connect using TLS
+~~~~~~~~~~~~~~~~~
 
 For ``tls`` connections:
-
-- You can use the 
-  :manual:`tls option </reference/connection-string/#urioption.tls>` to 
-  set ``tls=true`` in the 
-  :manual:`connection string </reference/connection-string>`:
-
-  .. example::
-
-     To enable ``tls`` in the connection string:
-
-     .. code-block:: sh
-
-        mongosh "mongodb://mongodb0.example.com:28015/?tls=true"
 
 - If you are using the :manual:`DNS Seedlist Connection Format 
   </reference/connection-string/#dns-seedlist-connection-format>`, the 
@@ -161,6 +154,19 @@ For ``tls`` connections:
      .. code-block:: sh
 
         mongosh "mongodb+srv://server.example.com/"
+
+- You can use the 
+  :manual:`tls option </reference/connection-string/#urioption.tls>` to 
+  set ``tls=true`` in the 
+  :manual:`connection string </reference/connection-string>`:
+
+  .. example::
+
+     To enable ``tls`` in the connection string:
+
+     .. code-block:: sh
+
+        mongosh "mongodb://mongodb0.example.com:28015/?tls=true"
 
 - You can specify the ``--tls`` command-line option.
   
--- a/source/connect.txt
+++ b/source/connect.txt
@@ -12,8 +12,6 @@ Connect to a Deployment
    :depth: 2
    :class: singlecol
 
-Prerequisites
-~~~~~~~~~~~~~
 
 Ensure that a MongoDB deployment is running before attempting to 
 connect to it with ``mongosh``.
--- a/source/connect.txt
+++ b/source/connect.txt
@@ -16,11 +16,8 @@ Connect to a Deployment
 Ensure that a MongoDB deployment is running before attempting to 
 connect to it with ``mongosh``.
 
-Connect to an Instance
-----------------------
-
 Local MongoDB Instance on Default Port
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+--------------------------------------
 
 You can run ``mongosh`` without any command-line options
 to connect to a MongoDB instance running on your **localhost** with 
@@ -31,7 +28,7 @@ to connect to a MongoDB instance running on your **localhost** with
    mongosh
 
 Local MongoDB Instance on a Non-Default Port
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+--------------------------------------------
 
 To specify a port to connect to on localhost, use the 
 :manual:`connection string </reference/connection-string>` format. For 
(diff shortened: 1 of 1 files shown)


source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
 .. contents:: On this page
    :local:
    :backlinks: none
-   :depth: 1
+   :depth: 2
    :class: singlecol
+
+
+Ensure that a MongoDB deployment is running before attempting to
```
jeff-allen-mongo, line 16: [suggest] This is easy to scroll through / miss. Possibly make this an `important` or add a `Prerequisite` heading?

source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
...
+
+
+Ensure that a MongoDB deployment is running before attempting to 
+connect to it with ``mongosh``. The connection method you use depends 
+on your deployment type.
+
+Local MongoDB Instance on Default Port
+--------------------------------------
+
+You can run ``mongosh`` without any command-line options
```
jeff-allen-mongo, line 23: [suggest] Remove "You can" and just start with "Run". More direct.

source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
...
+.. code-block:: sh
+
+   mongosh
+
+Local MongoDB Instance on a Non-Default Port
+--------------------------------------------
+
+To specify a port to connect to on localhost, use the 
+:manual:`connection string </reference/connection-string>` format. For 
+example, to connect to a MongoDB instance running on localhost with a
```
jeff-allen-mongo, line 36: Consider using an `example` block instead of "For example..."

source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
...
+non-default port 28015:
+
+.. code-block:: sh
+
+   mongosh "localhost:28015"
+
+MongoDB Instance on a Remote Host
+---------------------------------
+
+To specify a remote host and port, use a :manual:`connection string
```
jeff-allen-mongo, line 46: [optional] We can maybe highlight Atlas a bit more here. Consider one of:

- Making the title `MongoDB Instance on a Remote Host (Atlas)`

- Making the opening the sentence "To specify a remote host and port, such as an Atlas cluster, use a ..."

source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
...
+</reference/connection-string>`. For example, to connect to a MongoDB
```
jeff-allen-mongo, line 47: Again, consider example blocks.

source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
...
+instance running on a remote host machine on port 28015:
```
jeff-allen-mongo, line 48: Can we make this example Atlas specific? i.e., 

For example, to connect to a MongoDB Atlas cluster, you might use the following connection string:

source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
...
+  .. code-block:: sh
+
+     mongosh "mongodb://alice:<password>@mongodb0.examples.com:28015/?authSource=admin"
+
+- You can use the ``--username``, ``--password``, and
+  ``--authenticationDatabase`` command-line options. For example, to
+  connect to a remote MongoDB instance and authenticate against the 
+  ``admin`` database as user ``alice``:
+
+  .. code-block:: sh
```
jeff-allen-mongo, line 82: Can we adjust this example to discourage users from typing their passwords in plaintext? Let's instead direct them to just specify --username and say they will be prompted for a password.

Maybe then use the note below to call out the --password option, if they want to use that.

source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
...
+  can include the ``+srv`` modifier in your connection string:
+
+  .. code-block:: sh
+
+     mongosh "mongodb+srv://server.example.com/"
+
+  .. note::
+     Using the ``+srv`` connection string modifier automatically sets 
+     the 
+     :manual:`tls option </reference/connection-string/#urioption.tls>`
```
jeff-allen-mongo, line 114: Just a note, no action required: when we have the shell options in the mongosh repo linkable, we should link there instead of the manual. Need https://jira.mongodb.org/browse/DOP-1145 resolved.

source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
...
+To connect to a replica set:
+
+- You can specify the replica set name and members in the
+  :manual:`connection string </reference/connection-string>`:
+
+  .. code-block:: sh
+
+     mongosh "mongodb://mongodb0.example.com.local:27017,mongodb1.example.com.local:27017,mongodb2.example.com.local:27017/?replicaSet=replA"
+
+- If you are using the :manual:`DNS Seedlist Connection Format
```
jeff-allen-mongo, line 103: I'm just wondering if the DNS Seedlist Format should come first. I don't know or feel strongly about this, but I'm wondering if because it's generally simpler we should be pushing users more toward it. Just something to think about.

source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
...
+- If you are using the :manual:`DNS Seedlist Connection Format 
+  </reference/connection-string/#dns-seedlist-connection-format>`, the 
+  ``+srv`` connection string modifier automatically sets
+  the ``tls`` option to ``true`` for the connection:
+
+  .. code-block:: sh
+
+     mongosh "mongodb+srv://server.example.com/"
+
+- You can specify the ``--tls`` command-line option.
```
jeff-allen-mongo, line 141: Does this work for both the DNS format and non-DNS connection string? If so, we might want to explicitly say so.

source/connect.txt
zach-carr: Will do, that makes sense - I'm reading this as remove the connection string option entirely (since you can't connect without providing the password there, too). I'll save that copy in case that's not what you had in mind.

source/connect.txt
zach-carr: The DNS format automatically enables tls - I'll add a note.

source/connect.txt
```diff
@@ -9,5 +9,147 @@ Connect to a Deployment
...
+To connect to a MongoDB instance requires authentication:
+
+- You can specify the username, password, and authentication database 
+  in the :manual:`connection string </reference/connection-string>`. 
+  For example, to connect to a remote MongoDB instance and
+  authenticate against the ``admin`` database as user ``alice``:
+
+  .. code-block:: sh
+
+     mongosh "mongodb://alice:<password>@mongodb0.examples.com:28015/?authSource=admin"
```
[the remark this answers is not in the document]
zach-carr: It does not - I removed the option to provide credentials in the connection string entirely.

source/connect.txt
zach-carr: IMO this is the best example because it's the most basic, and there's a note below about connecting to Atlas instead. I worry that if we use the Atlas example as the default it confuses things conceptually a little bit. Not a strong opinion so happy to go with this if you still feel it's more common/the way of the future to use Atlas.

source/connect.txt
zach-carr: Makes sense to me, done.

source/connect.txt
zach-carr: I went back on this because I decided not to change the example to an Atlas example. Without this being at Atlas example it read strangely to me. It's still a possible that this should be Atlas focused, so if that's what we decided I'll make both changes together.

zach-carr:
Thanks @jeff-allen-mongo! This is ready for another look. Things that might be worth noting:
1. I did not take the suggestion to make the remote host an Atlas example, but am open if you feel strongly.
2. I removed the option to pass credentials in the connection string, which may or may not be what you had in mind. I can add it back real fast if it wasn't. 

@mmarcon, would you please review this page as well whenever you get some time? Thank you!

source/connect.txt
```diff
@@ -9,5 +9,174 @@ Connect to a Deployment
...
    :local:
    :backlinks: none
-   :depth: 1
+   :depth: 2
    :class: singlecol
+
+
+.. important::
+
+   Ensure that a MongoDB deployment is running before attempting to
```
jeff-allen-mongo, line 18: [nit] Maybe something like "Ensure that you have a running MongoDB deployment before attempting to connect with ``mognosh``" ?

Reason being, as written it sounds like "ensure that somewhere in the world a MongoDB deployment is running before you connect."

source/connect.txt
```diff
@@ -9,5 +9,174 @@ Connect to a Deployment
...
+Local MongoDB Instance on Default Port
+--------------------------------------
+
+Run ``mongosh`` without any command-line options
+to connect to a MongoDB instance running on your **localhost** with 
+**default port** 27017:
+
+.. code-block:: sh
+
+   mongosh
```
jeff-allen-mongo, line 31: [optional] possibly add a second code-block after this one like, "This is equivalent to running the following command:

.. code-block:: sh

      mongosh "localhost:27017"

source/connect.txt
```diff
@@ -9,5 +9,174 @@ Connect to a Deployment
...
+
+  .. example::
+
+     To enable ``tls`` in the connection string:
+
+     .. code-block:: sh
+
+        mongosh "mongodb://mongodb0.example.com:28015/?tls=true"
+
+- If you are using the :manual:`DNS Seedlist Connection Format
```
jeff-allen-mongo, line 151: Do we want to move this DNS example the top to match the other section where we made DNS first?

source/connect.txt
```diff
@@ -9,5 +9,174 @@ Connect to a Deployment
...
+
+   If your remote host is a |service-fullname| cluster, you can copy 
+   your connection string from the |service| UI. To learn more, see 
+   :atlas:`Connect to a Cluster <connect-to-cluster/#use-the-connect-dialog-to-connect-to-your-cluster>`.
+
+Connection Considerations
+-------------------------
+
+Authentication
+~~~~~~~~~~~~~~
```
jeff-allen-mongo, line 73: I'm wondering if this should be under Considerations, or its own top-level heading like "Connect with Authentication". Thoughts?

source/connect.txt
zach-carr: Yes, good call

source/connect.txt
zach-carr: I think that it's logically grouped as kind of a 'connection modifier' to be applied on top of the connection string format of whatever instance type you're trying to connect to. That said, I agree that auth is more important than a 'consideration' - thinking about better names for this whole section.

jeff-allen-mongo:
@mmarcon, I'm merging these docs now so we can publish for the devhub post. If there are any changes to make please let us know and we can do so post-publish.

zach-carr:
I've created a ticket to add --port and --host options: https://jira.mongodb.org/browse/DOCSP-10666

I think it makes sense to do tech review for this whole page as part of that upcoming PR so we're not spending time on old content, so I'm closing the ticket for this PR.

PR #15, opened 2020-06-05
(DOCSP-10539): Install tweaks for Homebrew

jeff-allen-mongo:
Scope of Changes:

- Moved Install Instructions to their own page. Motivation: This section was getting long and the home page was kind of losing its meaning. Instead, I'm trying to keep the scope home page to high-level, important features of the MongoDB Shell. This also provides symmetry with the rest of our docs where most often Install gets its own page.

- Added instructions for install with Homebrew.

- Tweaked instructions for package install. Specifically:

  - Make mongosh executable
  - Remove mongosh from quarantine

- Added a "Next Steps" section pointing users to the connection page.

Staging:

[Install the MongoDB Shell](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker/DOCSP-10539/install)

### DIFF
--- a/source/includes/steps-install-shell-base.yaml
+++ b/source/includes/steps-install-shell-base.yaml
@@ -34,7 +34,7 @@ content: |
 
    .. code-block:: sh
 
-     chmod +x /path/to/mongosh
+      chmod +x /path/to/mongosh
 ---
 title: "Add the MongoDB Shell binary to your ``PATH`` environment
    variable."
--- a/source/includes/steps-install-shell-base.yaml
+++ b/source/includes/steps-install-shell-base.yaml
@@ -23,12 +23,18 @@ ref: extract-archive
 level: 4
 content: |
    
-   .. code-block::
+   .. code-block:: sh
 
       tar -zxvf path/to/archive
 
-   Skip this step if your web browser automatically unzips the file as
-   part of the download.
+   If your web browser automatically extracts the archive as
+   part of the download or you extract the archive without the
+   ``tar`` command, you may need to run the following command to
+   make ``mongosh`` executable:
+
+   .. code-block:: sh
+
+     chmod +x /path/to/mongosh
 ---
 title: "Add the MongoDB Shell binary to your ``PATH`` environment
    variable."
--- a/source/includes/steps-install-shell-macos-homebrew.yaml
+++ b/source/includes/steps-install-shell-macos-homebrew.yaml
@@ -4,12 +4,8 @@ ref: install-homebrew
 level: 4
 content: |
 
-   Run the following command to install the
-   `Homebrew <https://brew.sh/>`__ package manager:
-
-   .. code-block:: sh
-  
-      /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+   Refer to the `Homebrew <https://brew.sh/>`__ website
+   for instructions to install Homebrew on macOS.
 ---
 title: "Tap the MongoDB Homebrew Tap."
 ref: tab-mongodb
--- a/source/includes/steps-install-shell-macos-homebrew.yaml
+++ b/source/includes/steps-install-shell-macos-homebrew.yaml
@@ -1,0 +1,37 @@
+---
+title: "Install Homebrew."
+ref: install-homebrew
+level: 4
+content: |
+
+   Run the following command to install the
+   `Homebrew <https://brew.sh/>`__ package manager:
+
+   .. code-block:: sh
+  
+      /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+---
+title: "Tap the MongoDB Homebrew Tap."
+ref: tab-mongodb
+level: 4
+content: |
+
+   Issue the following command from the terminal to tap the official
+   `MongoDB Homebrew Tap <https://github.com/mongodb/homebrew-brew>`__:
+
+   .. code-block:: sh
+
+      brew tap mongodb/brew
+---
+title: "Install the ``mongosh`` package."
+ref: install-mongosh-package
+level: 4
+content: |
+
+   Issue the following command from the terminal to install the
+   ``mongosh`` package:
+
+   .. code-block:: sh
+
+      brew install mongosh
+...
--- a/source/includes/steps-install-shell-macos-homebrew.yaml
+++ b/source/includes/steps-install-shell-macos-homebrew.yaml
@@ -5,7 +5,7 @@ level: 4
 content: |
 
    Refer to the `Homebrew <https://brew.sh/>`__ website
-   for instructions to install Homebrew on macOS.
+   for the steps to install Homebrew on macOS.
 ---
 title: "Tap the MongoDB Homebrew Tap."
 ref: tab-mongodb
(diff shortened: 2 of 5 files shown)


source/includes/steps-install-shell-base.yaml
```diff
@@ -23,12 +23,18 @@ ref: extract-archive
...
-   Skip this step if your web browser automatically unzips the file as
-   part of the download.
+   If your web browser automatically extracts the archive as
+   part of the download or you extract the archive without the
+   ``tar`` command, you may need to run the following command to
+   make ``mongosh`` executable:
+
+   .. code-block:: sh
+
+     chmod +x /path/to/mongosh
```
jwilliams-mongo, line 37:
```suggestion
      chmod +x /path/to/mongosh
```
seems like indentation is off -- doesn't seem to affect the rendering in nextgen, however.

source/includes/steps-install-shell-macos-homebrew.yaml
```diff
@@ -0,0 +1,37 @@
...
+ref: install-homebrew
+level: 4
+content: |
+
+   Run the following command to install the
+   `Homebrew <https://brew.sh/>`__ package manager:
+
+   .. code-block:: sh
+  
+      /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```
andf-mongodb, line 12: We should not be providing this step, but should instead link to the Homebrew instructions themselves:
- If homebrew changes this syntax, we have to remember to change this to match
- If a user is following MDB instructions, but runs into an issue specifically with the Homebrew install step that we provide, they will expect us to provide troubleshooting steps for this step. If we instead link to the Homebrew install steps, users are on the Homebrew page when they run into their issues, and are in the right place to ask for support, or click sidebar links to the Homebrew troubleshooting guide (i.e. read https://docs.brew.sh/ not https://docs.mongodb.com)

jeff-allen-mongo:
adding @andf-mongodb to review who mentioned they have some notes on Homebrew install.

source/includes/steps-install-shell-base.yaml
jeff-allen-mongo: Updating.

source/includes/steps-install-shell-macos.yaml
```diff
@@ -23,5 +23,17 @@ source:
...
+level: 4
+content: |
+   macOS may prevent ``mongosh`` from running after installation. If
+   you receive a security error when starting ``mongosh`` indicating
+   that the developer could not be identified or verified, run
+   the following command:
+
+   .. code-block:: sh
 
+      xattr -r -d com.apple.quarantine /path/to/mongosh
```
andf-mongodb, line 38: - Careful with this. Apple changes how it quarantines executables every version of macOS since it started doing this in 10.13. This command here works for me on 10.14, but I wonder about 10.13, and the future 10.16 (I'm assuming it worked for you on your 10.15 workstation).

- I use this copy for the install docs:
https://github.com/mongodb/docs/blob/8dc743535997ab1e68ba459f0ad2a18fe631f7a1/source/tutorial/install-mongodb-on-os-x.txt#L170
It's big, bulky, and I have to include it twice (once for `mongod` once for `mongo`), but it's GUI, familiar, and the same between versions. I don't want you to use this whole bulky copy from the install doc, it really is ugly & I hate it, but maybe something much shorter.

- the `-r` flag makes the command recursive, according to the man page. There's nothing to recurse under the `mongosh` executable itself. However, removing that flag made the command error for me, returning it restored functionality. ??

- can you run this command as a non-administrator? Internet search inconclusive.

source/install.txt
```diff
@@ -0,0 +1,80 @@
...
+      .. _macos-install-homebrew:
+
+      Install with Homebrew
+      ~~~~~~~~~~~~~~~~~~~~~
+
+      .. important::
+
+         To install ``mongosh`` with Homebrew on macOS, you must have a
+         minimum `node.JS <https://nodejs.org/en/download/>`__ version
+         of ``12.0.0``.
```
andf-mongodb, line 51: - Homebrew itself also has requirements: https://docs.brew.sh/Installation
- Not sure how much you want to include. If you direct users to the Homebrew install guide that lists these requirements, I think you don't need to list them here.

source/includes/steps-install-shell-macos-homebrew.yaml
jeff-allen-mongo: Agreed. Updating.

source/includes/steps-install-shell-macos.yaml
jeff-allen-mongo: - I saw your copy used on the main install docs. I was trying to avoid the bulky copy but you bring up a good point about this differing from version to version. Despite it being bulky, I can't think of a much clearer way to outline these actions. Still waiting on single-sourcing across multiple repos 🙃 

- Re `-r` flag: Weird right?? @mmarcon not sure if you know anything about this behavior?

- @mmarcon dropped a JIRA comment stating:

  > Note that sudo is only required for xattr if the binary is saved in a location where the user has no permissions. For example, I have it my ~/.bin folder and I could run the command without sudo.

source/install.txt
jeff-allen-mongo: Agreed. Added a specific blurb to this which links to that page.

source/includes/steps-install-shell-macos-homebrew.yaml
```diff
@@ -0,0 +1,33 @@
+---
+title: "Install Homebrew."
+ref: install-homebrew
+level: 4
+content: |
+
+   Refer to the `Homebrew <https://brew.sh/>`__ website
+   for instructions to install Homebrew on macOS.
```
andf-mongodb, line 8:
```suggestion
   for the steps to install Homebrew on macOS.
```

or "for the installation instructions for Homebrew  ... "

source/includes/steps-install-shell-macos.yaml
```diff
@@ -23,5 +23,21 @@ source:
...
+   you receive a security error when starting ``mongosh`` indicating
+   that the developer could not be identified or verified, perform
+   the following actions:
+
+   a. Open *System Preferences*.
+   
+   #. Select the *Security and Privacy* pane.
 
+   #. Under the *General* tab, click the button to the right of the
+      message about ``mongod``, labelled either :guilabel:`Open Anyway`
```
andf-mongodb, line 41:
```suggestion
      message about ``mongosh``, labelled either :guilabel:`Open Anyway`
```

source/includes/steps-install-shell-macos.yaml
andf-mongodb: Good sleuthing. I think the copy you used to walk users through this process is perfect. And a bit less bulkier! Nice work.

source/includes/steps-install-shell-macos.yaml
jeff-allen-mongo: Oops bad copy-paste. Thanks for flagging.

source/includes/steps-install-shell-macos-homebrew.yaml
jeff-allen-mongo: Agreed. Updating.

jeff-allen-mongo:
@mmarcon if you have a chance to ook at this before tomorrow morning that would be great. If not we can merge and tweak post-merge.

jeff-allen-mongo:
Hey @mmarcon, our platform team is publishing all of our docs this evening. I'm going to merge these updates in at around 4pm EST so they can go live. If there is anything we need to tweak here I'm happy to do so post-merge.

jeff-allen-mongo:
Merging now.

PR #23, opened 2020-07-15
(DOCSP-11251): Clarify that MongoDB Shell and mongosh are synonymous

jeff-allen-mongo:
Staged: https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-11251/

### DIFF
--- a/source/includes/admonitions/fact-mdb-shell-beta.rst
+++ b/source/includes/admonitions/fact-mdb-shell-beta.rst
@@ -1,6 +1,6 @@
 .. admonition:: Beta
    :class: note
 
-   |mdb-shell| is currently available as a **Beta** release. The
-   product, its features, and the corresponding documentation may
-   change during the Beta stage.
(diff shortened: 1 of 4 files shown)


jeff-allen-mongo:
Hey @mmarcon,

I'm opening this PR to better clarify that MongoDB Shell and `mongosh` are synonymous in the sense the `mongosh` is the binary used for the new shell. For context, we're adding some content to the [mongo shell](https://docs.mongodb.com/manual/mongo/) docs that calls out the new shell and highlights the differences between the two. We don't want to be linking from a page that says `mongo shell` to a page that says `MongoDB Shell` because that seems like an easy way to confuse our readers. So, we're referring to the new shell as `mongosh` on the legacy mongo shell side and making these updates here to make it clear that `mongosh` == MongoDB shell.

I hope this all makes sense. Feel free to ping or set up a call if you want to discuss this in greater detail.

PR #27, opened 2020-08-10
(DOCSP-11553): Add user management methods

jeff-allen-mongo:
Just to note for review, I'll also be updating the `passwordPrompt()` docs in the manual to reflect the mongosh limitation (separate ticket).

Staged: [User Management Methods](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-11553-new/reference/methods#user-management-methods)

### DIFF
--- a/source/connect.txt
+++ b/source/connect.txt
@@ -212,13 +212,13 @@ To connect to a specific database, specify a database in your
 <https://docs.mongodb.com/manual/reference/connection-string/>`__. If
 you do not specify a database in your URI path, you connect to a database named ``test``.
 
-   .. example::
+.. example::
 
-     The following connection string URI connects to database ``db1``.
+  The following connection string URI connects to database ``db1``.
 
-     .. code-block:: sh
+  .. code-block:: sh
 
-         mongosh "mongodb://localhost:27017/db1" 
+      mongosh "mongodb://localhost:27017/db1" 
 
 Connect to a Different Deployment
 ---------------------------------
--- a/source/reference/methods.txt
+++ b/source/reference/methods.txt
@@ -407,6 +407,72 @@ Database Methods
 
      - Provides access to the specified database.
 
+   * - :method:`db.logout()`
+
+     - Ends an authenticated session.
+
    * - :method:`db.runCommand()`
 
      - Runs a :manual:`database command </reference/command>`.
+
+User Management Methods
+-----------------------
+
+.. important::
+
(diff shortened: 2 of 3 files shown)


source/connect.txt
```diff
@@ -212,13 +212,13 @@ To connect to a specific database, specify a database in your
 <https://docs.mongodb.com/manual/reference/connection-string/>`__. If
 you do not specify a database in your URI path, you connect to a database named ``test``.
 
-   .. example::
```
jeff-allen-mongo, line 215: This is to fix a build warning. Not directly related to this ticket.

source/write-scripts.txt
```diff
@@ -48,7 +48,7 @@ Consider a MongoDB instance running on localhost on the default port.
 
    - Instantiates a new connection to the instance, and
    - Sets the global ``db`` variable to ``myDatabase`` using the
-     :method:`getDB()` method. 
+     :method:`Mongo.getDB()` method.
```
jeff-allen-mongo, line 51: This is to fix a build warning. Not directly related to this ticket.

source/reference/methods.txt
```diff
@@ -444,6 +444,72 @@ Database Methods
...
+   - :method:`db.auth()`
+   - :method:`db.changeUserPassword()`
+   - :method:`db.createUser()`
+   - :method:`db.updateUser()`
+
+.. list-table::
+   :widths: 30 70
+   :header-rows: 1
+
+   * - Name
```
jeff-allen-mongo, line 473: Doh this should be `Method`, not `Name`. Will fix in next patch.

source/reference/methods.txt
```diff
@@ -444,6 +444,72 @@ Database Methods
...
      - Runs a :manual:`database command </reference/command>`.
+
+User Management Methods
+-----------------------
+
+.. important::
+
+   The :method:`passwordPrompt()` method is currently not
+   supported in ``mongosh``. This limitation affects behavior
+   for the following methods:
```
nathan-contino-mongo, line 462: I'm guessing that "affects behavior" means that the user has to enter the password as a parameter instead of using the password prompt -- can we make that clearer here? Otherwise I think it's hard to really draw any information out of this admonition unless you're already very familiar with `passwordPrompt()`.

jeff-allen-mongo: Good call! Updated with more context. Thanks Nate

PR #33, opened 2020-08-17
(DOCSP-11439): BSON class help

jeff-allen-mongo:
JIRA: [DOCSP-11439](https://jira.mongodb.org/browse/DOCSP-11439?)

Staged: [BSON class help](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-11439/reference/access-mdb-shell-help#bson-class-help)

Need to follow up with tech review because not all .help() methods added in [the mongosh commit](https://github.com/mongodb-js/mongosh/commit/96a730fc05e4c226e9c365e971751bf0f362af1b) seem to be working.

The following classes' help methods don't work in mongosh:

- `bsonsize`
- `MD5`
- `UUID`
- `HexData`

They return `TypeError: <className>.help is not a function`

### DIFF
--- a/source/reference/access-mdb-shell-help.txt
+++ b/source/reference/access-mdb-shell-help.txt
@@ -18,9 +18,8 @@ information.
 
 .. tip::
 
-   When accessing :ref:`database <mdb-shell-help-db>` and
-   :ref:`collection <mdb-shell-help-collection>` help in ``mongosh``,
-   you can use the ``.help()`` and ``.help`` syntaxes interchangeably.
+   When accessing help in ``mongosh``, you can use the ``.help()`` and
+   ``.help`` syntaxes interchangeably.
 
 .. _mdb-shell-help-command-line:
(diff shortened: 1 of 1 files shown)


source/reference/access-mdb-shell-help.txt
```diff
@@ -242,3 +241,47 @@ Show Additional Usage Details for a Collection Method
...
+   ObjectId.help()
+
+``mongosh`` returns the following output:
+
+.. code-block:: javascript
+   :copyable: false
+
+   The ObjectId BSON Class:
+
+
```
jwilliams-mongo, line 272: is there an extra return here?

jeff-allen-mongo: This is what the output looks like in the shell, but I think we can safely remove extra returns in the docs. Done.

PR #34, opened 2020-08-18
DOCSP-11793: [MONGOSH] Support Bulk operation methods

steveren:
[Jira](https://jira.mongodb.org/browse/DOCSP-11793)
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-11793/reference/methods#bulk-operation-methods)

### DIFF
--- a/source/reference/methods.txt
+++ b/source/reference/methods.txt
@@ -264,17 +264,17 @@ Bulk Operation Methods
 
    * - :method:`db.collection.initializeOrderedBulkOp()`
 
-     - Initializes and returns a new :method:`Bulk <db.collection.Bulk()>` operations
-       builder for a collection. The builder constructs an ordered list of
-       write operations that MongoDB executes in bulk.
+     - Initializes and returns a new :method:`Bulk()` operations builder
+       for a collection. The builder constructs an ordered list of write
+       operations that MongoDB executes in bulk.
 
    * - :method:`db.collection.initializeUnorderedBulkOp()`
 
-     - Initializes and returns a new :method:`db.collection.Bulk()` operations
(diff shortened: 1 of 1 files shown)


source/reference/methods.txt
```diff
@@ -251,10 +251,111 @@ Collection Methods
...
+   * - :method:`Bulk.execute()`
+
+     - Executes the list of operations built by the :method:`Bulk()`
+       operations builder.
+
+   * - :method:`Bulk.find()`
+
+     - Specifies a query condition for an update or a remove operation.
+
+   * - :method:`Bulk.find.arrayFilters()`
```
jeff-allen-mongo, line 293: The ticket says that `arrayFilters` and `collation` are blocked. Do they work in mongosh? If not, I think we should hold off on documenting them here.

steveren: Whoops, my mistake, removed those two methods.

source/reference/methods.txt
```diff
@@ -251,10 +251,99 @@ Collection Methods
...
+
+.. list-table::
+   :widths: 30 70
+   :header-rows: 1
+
+   * - Method
+
+     - Description
+
+   * - :method:`db.collection.initializeOrderedBulkOp()`
```
jeff-allen-mongo, line 265: Should these go after the `Bulk...` methods if we're going for alphabetical?

steveren: I used the same ordering that they do in the [manual](https://docs.mongodb.com/manual/reference/method/js-bulk/). I would imagine it's because all the other methods are basically sub-methods to those top two; you can't use the sub-methods without first calling one of the top-level methods.

PR #49, opened 2020-09-16
(DOCSP-12243): Add Run Commands page and some command behaviors

jeff-allen-mongo:
New updates for behavior when pressing ctrl+c in the shell, as well as the breaking change (compared to legacy mongo) for Command Exceptions. Moved the CRUD and Agg Pipeline docs to sub-pages of this new page.

Staged: [Run Commands in mongosh](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-12243/run-commands)

Also updated behavior in Changelog. [Staging - v3.0.1 Behavior Updates](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-12243/changelog#v0-3-1)

### DIFF
--- a/source/run-commands.txt
+++ b/source/run-commands.txt
@@ -1,0 +1,54 @@
+.. _run-commands:
+
+===========================
+Run Commands in ``mongosh``
+===========================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+.. include:: /includes/admonitions/fact-mdb-shell-beta.rst
+
+For documentation of basic MongoDB operations in ``mongosh``, see:
+
+- :doc:`/crud`
+
+- :doc:`/run-agg-pipelines`
+
+- :doc:`/reference/methods`
+
+Terminate a Running Command
+---------------------------
+
+To terminate a running command or query in ``mongosh``,
+press ``Ctrl + C``.
+
+``Ctrl + C`` terminates the process in the shell, but does not
+terminate the process on the MongoDB server. This behavior
+differs from the :method:`db.killOp()` method, which terminates
+commands on the server.
+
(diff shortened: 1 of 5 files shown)


source/run-commands.txt
```diff
@@ -0,0 +1,57 @@
...
+press ``Ctrl + C``.
+
+.. note::
+
+   ``Ctrl+C`` terminates the process in the shell, but does not
+   terminate the process on the MongoDB server.
+
+.. seealso::
+
+   :method:`db.killOp()`
```
aherlihy, line 38: Perhaps it might make sense to specify that this does _not_ use killOp and that killOp can be used to kill commands that are running on the server?

jeff-allen-mongo:
Thanks @aherlihy, you bring up a good point. I tweaked the copy to better clarify the difference in behavior between ctrl+c and killOp()

source/run-commands.txt
```diff
@@ -0,0 +1,53 @@
...
+
+Command Exceptions
+------------------
+
+When running a helper method which wraps a database command (such as
+:method:`db.getCollectionNames`), if the server would return
+``{ ok: 0 }``, thereby indicating the method was not successful,
+``mongosh`` raises an exception.
+
+If you run the command using :method:`db.runCommand()` and the command
```
aherlihy, line 44: Turns out this is actually not true - the driver will return an error even for `db.command`. If we're okay with it always throwing, then we should just update the docs to indicate that it will return an error in any case that `ok: 0`. If not OK then we have to make some code changes to make this statement true

jeff-allen-mongo: Tweaked the copy both here and in the changelog for the clarified behavior. Thanks!

aherlihy (approved):

PR #56, opened 2020-09-25
(DOCSP-11649): Update install instructions for .deb and .rpm

jeff-allen-mongo:
Added tabs to the Linux tab for each package / Linux distro supported. If you think there's a better way to structure this, happy to discuss.

Staged: [Install mongosh](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-11649/install#install-mongosh)

### DIFF
--- a/source/includes/steps-install-shell-base.yaml
+++ b/source/includes/steps-install-shell-base.yaml
@@ -1,20 +1,14 @@
 ---
-title: "Navigate to the MongoDB Download Center."
+title: "Open the MongoDB Shell download page."
 ref: navigate-dlc
 level: 4
 content: |
 
   Open the `MongoDB Download Center
-  <https://www.mongodb.com/download-center/community?jmp=docs>`__.
+  <https://www.mongodb.com/try/download/shell?jmp=docs>`__.
 ---
-title: "Open the |mdb-shell| Download Page."
-ref: open-mdb-shell-page
-level: 4
-content: |
-
-   Click :guilabel:`Tools`, then select :guilabel:`Shell`.
----
-title: "Download the installation archive for your operating system."
+title: "Download the ``mongosh`` installation archive for your
+  operating system."
 ref: download-archive
 level: 4 
 ---
@@ -22,10 +16,13 @@ title: "Extract the files from the downloaded archive."
 ref: extract-archive
 level: 4
 content: |
+
+   Run the following command from the directory containing the
+   ``mongosh`` ``.tgz`` package:
    
    .. code-block:: sh
 
-      tar -zxvf path/to/archive
+      tar -zxvf mongosh-{+version+}-linux.tgz
 
    If your web browser automatically extracts the archive as
    part of the download or you extract the archive without the
@@ -34,7 +31,7 @@ content: |
 
    .. code-block:: sh
 
-      chmod +x /path/to/mongosh
+      chmod +x mongosh
 ---
 title: "Add the ``mongosh`` binary to your ``PATH`` environment
    variable."
@@ -44,21 +41,20 @@ content: |
 
    You can either:
 
-   - Copy the ``mongosh`` binary into a directory listed in your ``PATH``
-     variable, such as ``/usr/local/bin`` (Update
-     ``/path/to/mongosh`` with your installation directory as
-     appropriate)
+   - Copy the ``mongosh`` binary into a directory listed in your
+     ``PATH`` variable, such as ``/usr/local/bin``. Run the following
+     command from the directory containing the ``mongosh`` binary:
 
      .. code-block:: sh
 
-       sudo cp /path/to/mongosh /usr/local/bin/
+       sudo cp mongosh /usr/local/bin/
 
    - Create a symbolic link to the ``mongosh`` binary from a directory
-     listed in your ``PATH`` variable, such as ``/usr/local/bin``
-     (Update ``/path/to/mongosh`` with your installation directory as
-     appropriate):
+     listed in your ``PATH`` variable, such as ``/usr/local/bin``.
+     Run the following command from the directory containing the
+     ``mongosh`` binary:
 
      .. code-block:: sh
 
-       sudo ln -s  /path/to/mongosh /usr/local/bin/
+       sudo ln -s mongosh /usr/local/bin/
 ...
--- a/source/includes/steps-install-shell-linux-deb.yaml
+++ b/source/includes/steps-install-shell-linux-deb.yaml
@@ -12,15 +12,10 @@ title: "Extract and install the ``.deb`` package."
 ref: extract-archive-linux-deb
 level: 4
 content: |
-  To install the downloaded ``.deb`` package, either:
+  To install the |mdb-shell|, run the following command from the
+  directory containing the ``mongosh`` ``.deb`` package:
 
-  - Double click your downloaded ``.deb`` package icon and follow
-    the prompts.
+  .. code-block:: sh
 
-  - Run the following command from the directory containing the
-    ``mongosh`` ``.deb`` package:
-
-    .. code-block:: sh
-
-       sudo dpkg --install mongosh_{+version+}_amd64.deb
+      sudo dpkg --install mongosh_{+version+}_amd64.deb
 ...
--- a/source/includes/steps-install-shell-linux-deb.yaml
+++ b/source/includes/steps-install-shell-linux-deb.yaml
@@ -4,15 +4,9 @@ source:
   ref: navigate-dlc
 ref: navigate-dlc-linux-deb
 ---
-source:
-  file: steps-install-shell-base.yaml
-  ref: open-mdb-shell-page
-ref: open-mdb-shell-page-linux-deb
----
-source:
-  file: steps-install-shell-base.yaml
-  ref: download-archive
-ref: download-archive-generic-linux
+title: "Download the Debian 64-bit ``.deb`` package."
+ref: download-archive-rhel
+level: 4 
 ---
 title: "Extract and install the ``.deb`` package."
 ref: extract-archive-linux-deb
@@ -23,9 +17,10 @@ content: |
   - Double click your downloaded ``.deb`` package icon and follow
     the prompts.
 
-  - Run the following command from the terminal:
+  - Run the following command from the directory containing the
+    ``mongosh`` ``.deb`` package:
 
     .. code-block:: sh
 
-       sudo dpkg --install path/to/mongosh_<version>_.deb
+       sudo dpkg --install mongosh_{+version+}_amd64.deb
 ...
--- a/source/includes/steps-install-shell-linux-deb.yaml
+++ b/source/includes/steps-install-shell-linux-deb.yaml
@@ -1,0 +1,31 @@
+---
+source:
+  file: steps-install-shell-base.yaml
+  ref: navigate-dlc
+ref: navigate-dlc-linux-deb
+---
+source:
+  file: steps-install-shell-base.yaml
+  ref: open-mdb-shell-page
+ref: open-mdb-shell-page-linux-deb
+---
+source:
+  file: steps-install-shell-base.yaml
+  ref: download-archive
+ref: download-archive-generic-linux
+---
+title: "Extract and install the ``.deb`` package."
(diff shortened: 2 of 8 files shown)


source/includes/steps-install-shell-linux-deb.yaml
```diff
@@ -0,0 +1,31 @@
...
+---
+source:
+  file: steps-install-shell-base.yaml
+  ref: open-mdb-shell-page
+ref: open-mdb-shell-page-linux-deb
+---
+source:
+  file: steps-install-shell-base.yaml
+  ref: download-archive
+ref: download-archive-generic-linux
```
andf-mongodb, line 15: • Step 1: You can direct users directly to tools DL page:

https://www.mongodb.com/try/download/tools

• Step 2: New DL center lists `mongosh` as "MongoDB Shell" not "Shell"

• Step 3: "DL for your platform" -- careful, Ubuntu is Linux as well. Since we have a dedicated tab for each, fair to just use what matches the tab:
- .tgz --> Download the "Linux 64-bit" ...
- .deb (Debian) --> Download the "Debian 64-bit" ...
- .rpm (RHEL)  --> "Download the 64-bit" ...

Possibly safe to skip the "64-bit", up to you.

• Step 3: However, this is rather hugely unfortunate, as the ".rpm (RHEL)" tab and "Rehat 64-bit" download will work for Amazon Linux and SUSE, likewise the ".deb (Debian)" tab and "Debian 64-bit" download will work for Ubuntu. Not sure how best to handle this. Certainly no need to bother with Ubuntu 20.04 vs Ubuntu 18.04 here, but I think adding something similar to "Use the .deb tab to install on the Debian and Ubuntu platforms" would be helpful.

source/install.txt
```diff
@@ -71,8 +71,26 @@ Procedure
...
+      Select your desired Linux installation package from the tabs
+      below:
+
+      .. tabs::
+      
+         .. tab:: .tgz
+            :tabid: tgz
       
-      .. include:: /includes/steps/install-shell-generic-linux.rst
+            .. include:: /includes/steps/install-shell-linux-generic.rst
```
andf-mongodb, line 83: The "Extract" step should list the TGZ file itself as well, as in:

`tar -zxvf mongosh-0.4.0-linux.tgz`

or 

`tar -zxvf /path/to/mongosh-0.4.0-linux.tgz`

andf-mongodb (commented):
Some things to think about for the installation steps.

Also, the `mongosh` [RPM](https://repo.mongodb.org/yum/redhat/8/mongodb-org/4.4/x86_64/RPMS/) and [DEB](https://repo.mongodb.org/apt/debian/dists/buster/mongodb-org/4.4/main/binary-amd64/) packages are available directly from our repo, meaning users can just `sudo yum install  mongosh` or `sudo apt-get install mongosh`. However, the DEB file is _not_ available in the Ubuntu 20.04 repo, for example, so I'm not sure if this is an oversight, or that Ubuntu is really not supported for mongosh. If so, disregard my comment to that effect inline in this review!

Anyways, some thing to think about -- let me know if you want to chat further.

source/includes/steps-install-shell-linux-deb.yaml
jeff-allen-mongo: 1. Done. Directed them right to shell download.
2. Blast, they changed it on us. Thanks for catching.
3. Yes, agreed this was a bit tricky. I like being explicit as you've suggested. Updating.
4. Added a clarifying section before the tabs.

**Question** Let me know if you would prefer the tab names to be the distributions (RHEL, Ubuntu/Debian, TGZ (ack that this is not a distribution)) rather than the package names. I went with package names since TGZ does not apply to a specific distro.

jeff-allen-mongo (commented):
Thanks for super helpful feedback Andrew. This is updated.

source/install.txt
```diff
@@ -71,8 +71,33 @@ Procedure
...
+
+      Select the appropriate tab based on your Linux distribution and
+      desired package:
+
+      - For TGZ tarball installation, click the ``.tgz`` tab.
+
+      - For installation on Ubuntu and Debian, click the ``.deb`` tab.
+
+      - For installation on :abbr:`RHEL (Red Hat Enterprise Linux)`,
+        click the ``.rpm`` tab.
```
andf-mongodb, line 83: • Nice. Only thought is that you can install the TGZ on Ubuntu and Debian and RHEL. To further differentiate, how about:

>Select the appropriate tab below based on your installation method:
> 
>- To install via the RPM package on RHEL, click ...
>- To install via the DEB package on Ubuntu and Debian, click ...
>- To install using the TGZ tarball, click ...

andf-mongodb: • "below" helps because they are now surrounded by tabs! This helps them to focus their attention on the right tabs for the next step.

• I generally leave TGZ steps to the end (i.e. as the third tab, here), but it's up to you. Ideally, most users will use the package manager steps.

source/includes/steps-install-shell-base.yaml
```diff
@@ -1,31 +1,28 @@
...
-title: "Open the |mdb-shell| Download Page."
-ref: open-mdb-shell-page
-level: 4
-content: |
-
-   Click :guilabel:`Tools`, then select :guilabel:`Shell`.
----
-title: "Download the installation archive for your operating system."
+title: "Download the ``mongosh`` installation archive for your
+  operating system."
```
andf-mongodb, line 11: • Thought: you could combine steps 1 & 2 here, like so:

> 1. Download mongosh
> Download the Red Hat 64-bit .rpm package from the <MongoDB Download Center>
>
> 1. Download mongosh
> Download the Debian 64-bit .deb package from the <MongoDB Download Center>
>
> 1. Download mongosh
> Download the Linux 64-bit .tgz package from the <MongoDB Download Center>

source/includes/steps-install-shell-base.yaml
```diff
@@ -1,31 +1,28 @@
...
 level: 4
 content: |
+
+   Run the following command from the directory containing the
+   ``mongosh`` ``.tgz`` package:
    
    .. code-block:: sh
 
-      tar -zxvf path/to/archive
+      tar -zxvf mongosh-{+version+}-linux.tgz
```
andf-mongodb, line 25: Perfect! This and the PATH changes look great!

source/includes/steps-install-shell-linux-rpm.yaml
```diff
@@ -0,0 +1,21 @@
...
+title: "Extract and install the ``.rpm`` package."
+ref: extract-archive-rpm
+level: 4
+content: |
+  To install the |mdb-shell|, run the following command from the
+  directory containing the ``mongosh`` ``.rpm`` package:
+
+  .. code-block:: sh
+
+     sudo yum install mongosh-{+version+}-x86_64.rpm
```
andf-mongodb, line 20:
```suggestion
     sudo yum install -y mongosh-{+version+}-x86_64.rpm
```

• Super minor. `yum` prompts you in the install: "are you really really sure you want to install??" For our purposes, we generally skip that prompt by passing `-y` to `yum`.

[Example in manual](https://docs.mongodb.com/manual/tutorial/install-mongodb-enterprise-on-red-hat/#install-mongodb-enterprise-version).

source/includes/steps-install-shell-linux-deb.yaml
```diff
@@ -0,0 +1,26 @@
...
+level: 4 
+---
+title: "Extract and install the ``.deb`` package."
+ref: extract-archive-linux-deb
+level: 4
+content: |
+  To install the downloaded ``.deb`` package, either:
+
+  - Double click your downloaded ``.deb`` package icon and follow
+    the prompts.
```
andf-mongodb, line 18: • This is a bit strange, as even DEB users who do have GUI / desktops would still likely run the command line step for this install. You can do the same thing for RPMs on RHEL (say on CentOS with GNOME), but that step is not present in the RPM steps. I think I'd just remove the GUI step, simplify this step a bit.

andf-mongodb (commented):
Nice changes! Especially love the untar and PATH updates, works really well.

Have a few technical thoughts, and one idea about combining steps. Only really important one is about differentiating between the three tabs, the rest are all up to you!

source/includes/steps-install-shell-linux-deb.yaml
andf-mongodb: I'd say you're absolutely right to go with package names over distro for exactly that reason. I think with the brief intro now preceding that second tier of tabs, we've added a lot of clarity to this user choice. Some thoughts about that intro inline.

source/includes/steps-install-shell-base.yaml
jeff-allen-mongo: I'm going to leave this as-is because this change would kind of muck up the single-sourcing for macOS and Windows, and I think this is ok enough where I don't want to spend the time to refactor.

jeff-allen-mongo:
@mmarcon can you take a pass at these updates when you have a chance? Thank you!

source/includes/steps-install-shell-linux-rpm.yaml
```diff
@@ -0,0 +1,21 @@
+---
+source:
+  file: steps-install-shell-base.yaml
+  ref: navigate-dlc
+ref: navigate-dlc-linux-rpm
+---
+title: "Download the Red Hat 64-bit ``.rpm`` package."
```
andf-mongodb, line 7: Minor: DL center lists this download entry as: 

> Redhat 64-bit

I could swear they changed that on us at some point!

andf-mongodb (approved):
LGTM mod on small fix

source/includes/steps-install-shell-linux-rpm.yaml
```diff
@@ -0,0 +1,21 @@
...
+title: "Extract and install the ``.rpm`` package."
+ref: extract-archive-rpm
+level: 4
+content: |
+  To install the |mdb-shell|, run the following command from the
+  directory containing the ``mongosh`` ``.rpm`` package:
+
+  .. code-block:: sh
+
+     sudo yum install -y mongosh-{+version+}-x86_64.rpm
```
mmarcon, line 20: Does this command work?

I don't have a RHEL-like box to try it but I _think_ this the command to install a package that is available in a repository. Does `yum` also allow installing RPMs directly? I thought one needs to `rpm -i mongosh-{+version+}-x86_64.rpm` but I might very well be wrong.

mmarcon (commented):
@jeff-allen-mongo when https://github.com/mongodb-js/mongosh/pull/346 is merged and has gone through Evergreen, the package should also be available for Ubuntu 20.04.

source/includes/steps-install-shell-linux-rpm.yaml
andf-mongodb: It's [an alias](https://serverfault.com/questions/940791/whats-the-difference-between-yum-install-local-path-and-yum-localinstall) to `localinstall` if passed an RPM package present in `cwd`.

`[vagrant@centos8 tmp]$ wget https://downloads.mongodb.com/compass/mongosh-0.4.0-x86_64.rpm
2020-09-29 14:42:48 (32.7 MB/s) - ‘mongosh-0.4.0-x86_64.rpm’ saved [33599000/33599000]
[vagrant@centos8 tmp]$ sudo yum install -y mongosh-0.4.0-x86_64.rpm
Total size: 32 M
Installed size: 136 M
Downloading Packages:
Running transaction check
Transaction check succeeded.
Running transaction test
Transaction test succeeded.
Running transaction
  Preparing        :                                                                                                                           1/1
  Installing       : mongosh-0.4.0-1.el8.x86_64                                                                                                1/1
  Running scriptlet: mongosh-0.4.0-1.el8.x86_64                                                                                                1/1
  Verifying        : mongosh-0.4.0-1.el8.x86_64                                                                                                1/1

Installed:
  mongosh-0.4.0-1.el8.x86_64

Complete!`

mmarcon: Ah, cool. Always happy to learn new stuff, thank you!

mmarcon (approved):

PR #58, opened 2020-10-13
DOCSP-12636 doc for Replica Set Status

kanchana-mongodb:
- [JIRA](https://jira.mongodb.org/browse/DOCSP-12636)
- [Stage](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-12636/reference/methods#replication-methods)

### DIFF
--- a/source/reference/methods.txt
+++ b/source/reference/methods.txt
@@ -888,12 +888,15 @@ Replication Methods
 
      - Description
  
-   * - ``rs.printSecondaryReplicationInfo()``
+   * - ``rs.printSecondaryReplicationInfo``
 
-     - Returns the status of the secondary members of the replica set. This 
-       is identical to the ``db.printSecondaryReplicationInfo()`` method. 
-       The server :method:`rs.printSlaveReplicationInfo()` method is referred 
-       to as ``rs.printSecondaryReplicationInfo()`` in ``mongosh``.
+     - Returns the status of the secondary members of the replica set. 
+     
+       This is identical to the ``db.printSecondaryReplicationInfo()`` 
+       method. This method returns the same output as the 
+       :method:`rs.printSlaveReplicationInfo()` method in the legacy 
+       :binary:`~mongo` shell. The :method:`rs.printSlaveReplicationInfo()` method is deprecated in ``mongosh``. Use 
+       ``rs.printSecondaryReplicationInfo()`` instead.
 
        The following is an example output from the 
        ``rs.printSecondaryReplicationInfo()`` method issued on a replica set with two secondary members:
@@ -913,12 +916,14 @@ Replication Methods
             replLag: '0 secs (0 hrs) behind the primary '
           }
 
-   * - ``db.printSecondaryReplicationInfo()``
+   * - ``db.printSecondaryReplicationInfo``
 
-     - Returns the status of the secondary members of the replica set. This 
-       is identical to the ``rs.printSecondaryReplicationInfo()`` method. 
-       The server :method:`db.printSlaveReplicationInfo()` method is referred 
-       to as ``db.printSecondaryReplicationInfo()`` in ``mongosh``.
+     - Returns the status of the secondary members of the replica set. 
+     
+       This is identical to the ``rs.printSecondaryReplicationInfo()`` method. 
+       This method returns the same output as the 
+       :method:`db.printSlaveReplicationInfo()` method. The :method:`db.printSlaveReplicationInfo()` method is deprecated in ``mongosh``. Use 
+       ``db.printSecondaryReplicationInfo()`` instead.
 
        The following is an example output from the 
        ``rs.printSecondaryReplicationInfo()`` method issued on a replica set with two secondary members:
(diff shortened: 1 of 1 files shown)


source/reference/methods.txt
```diff
@@ -876,3 +876,90 @@ Role Management Methods
...
+   * - Method
+
+     - Description
+ 
+   * - ``rs.printSecondaryReplicationInfo()``
+
+     - Returns the status of the secondary members of the replica set. This 
+       is identical to the ``db.printSecondaryReplicationInfo()`` method. 
+       The server :method:`rs.printSlaveReplicationInfo()` method is referred 
+       to as ``rs.printSecondaryReplicationInfo()`` in ``mongosh``.
```
jeff-allen-mongo, line 896: So I don't think this blurb is entirely correct. I don't think we want to contrast "server" methods and "mongosh" methods. :method:rs.printSlaveReplicationInfo() is a _legacy shell_ method, and we cant to contrast that with a _mongosh_ method. Also, 

Consider something like the following:

```
This is identical to the ``db.printSecondaryReplicationInfo()`` method. 
This method returns the same output as the :method:`rs.printSlaveReplicationInfo()`
method in the legacy binary:`~mongo` shell.
```

We may want to be more direct and explicitly mention something like

```
:method:`rs.printSlaveReplicationInfo()` is deprecated in mongosh.
Use :method:`rs.printSecondaryReplicationInfo()` instead.
```

source/reference/methods.txt
```diff
@@ -876,3 +876,90 @@ Role Management Methods
...
+   :widths: 30 70
+   :header-rows: 1
+
+   * - Method
+
+     - Description
+ 
+   * - ``rs.printSecondaryReplicationInfo()``
+
+     - Returns the status of the secondary members of the replica set. This
```
jeff-allen-mongo, line 893: Suggest putting a linebreak after the first sentence so the description of the method stands on its own and is easy to spot.

source/reference/methods.txt
```diff
@@ -876,3 +876,90 @@ Role Management Methods
...
+            replLag: '0 secs (0 hrs) behind the primary '
+          }
+          ---
+          source: rs3.example.net:27017
+          {
+            syncedTo: 'Tue Oct 13 2020 09:42:18 GMT-0700 (Pacific Daylight Time)',
+            replLag: '0 secs (0 hrs) behind the primary '
+          }
+
+   * - ``db.printSecondaryReplicationInfo()``
```
jeff-allen-mongo, line 916: See my comments above on ``rs.printSecondaryReplicationInfo()``.

source/reference/methods.txt
```diff
@@ -876,3 +876,90 @@ Role Management Methods
...
+          }
+
+   * - :method:`db.getReplicationInfo()`
+
+     - Returns the status of the replica set from the oplog data.
+
+   * - ``rs.isMaster()``
+
+     - Returns the replica set configuration, role, and status for 
+       the ``mongod`` instance that ``mongosh`` is connected to. The ``mongosh`` ``rs.isMaster()`` method is identical to the server :manual:`isMaster </reference/command/isMaster/>` command.
```
jeff-allen-mongo, line 948: [nits]

- Need a line break here.
- When linking to a command, you can do
```
:dbcommand:`isMaster`
```
instead of the `:manual:` link.

jeff-allen-mongo, line 948: I think rather than saying it is identical to the server isMaster command, it might be more accurate to say 

```
The ``mongosh`` ``rs.isMaster()`` method wraps the :dbcommand:`isMaster` command.
```

They're technically speaking not identical in behavior, although the output is the same.

jeff-allen-mongo (commented):
Thanks Kanchana! Leaving a few comments below just to clear up some technical points.

source/reference/methods.txt
```diff
@@ -876,3 +876,99 @@ Role Management Methods
...
+     - Description
+ 
+   * - ``rs.printSecondaryReplicationInfo``
+
+     - Returns the status of the secondary members of the replica set. 
+     
+       This is identical to the ``db.printSecondaryReplicationInfo()`` 
+       method. This method returns the same output as the 
+       :method:`rs.printSlaveReplicationInfo()` method in the legacy 
+       :binary:`~mongo` shell. The :method:`rs.printSlaveReplicationInfo()` method is deprecated in ``mongosh``. Use
```
jeff-allen-mongo, line 898: [nit] For the second instance of :method:`rs.printSlaveReplicationInfo()`, suggest just using monospace instead of the `method` link. I don't think it's super helpful to provide two links to the same location immediately after one another.

source/reference/methods.txt
```diff
@@ -876,3 +876,99 @@ Role Management Methods
...
+            replLag: '0 secs (0 hrs) behind the primary '
+          }
+
+   * - ``db.printSecondaryReplicationInfo``
+
+     - Returns the status of the secondary members of the replica set. 
+     
+       This is identical to the ``rs.printSecondaryReplicationInfo()`` method. 
+       This method returns the same output as the 
+       :method:`db.printSlaveReplicationInfo()` method. The :method:`db.printSlaveReplicationInfo()` method is deprecated in ``mongosh``. Use
```
jeff-allen-mongo, line 925: Ditto here for the removing the `method` link around the second instance of `db.printSlaveReplicationInfo()`. Also needs a line break here.

jeff-allen-mongo (approved):
LGTM with two nits. Thank you!

source/reference/methods.txt
aherlihy: I would also be wary of saying it 'return the same output' since it's formatted differently.

source/reference/methods.txt
```diff
@@ -876,3 +876,99 @@ Role Management Methods
...
+            syncedTo: 'Tue Oct 13 2020 09:42:18 GMT-0700 (Pacific Daylight Time)',
+            replLag: '0 secs (0 hrs) behind the primary '
+          }
+
+   * - ``db.printSecondaryReplicationInfo``
+
+     - Returns the status of the secondary members of the replica set. 
+     
+       This is identical to the ``rs.printSecondaryReplicationInfo()`` method. 
+       This method returns the same output as the
```
aherlihy, line 924: Again it's not exactly the same output as the old shell, since the formatting has been updated to be more readable.

aherlihy (approved):
Two comments but otherwise good!

PR #61, opened 2020-10-15
DOCSP-12664 Replica Set Management methods

kanchana-mongodb:
- [JIRA](https://jira.mongodb.org/browse/DOCSP-12664)
- [Stage](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-12664/reference/methods#replication-methods)

### DIFF
--- a/source/reference/methods.txt
+++ b/source/reference/methods.txt
@@ -887,34 +887,45 @@ Replication Methods
    * - Method
 
      - Description
- 
-   * - ``rs.printSecondaryReplicationInfo``
 
-     - Returns the status of the secondary members of the replica set. 
-     
-       This is identical to the ``db.printSecondaryReplicationInfo()`` 
-       method. This method's output is similar to the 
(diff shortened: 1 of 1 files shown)


source/reference/methods.txt
```diff
@@ -876,3 +876,31 @@ Role Management Methods
...
+   :header-rows: 1
+
+   * - Method
+
+     - Description
+
+   * - :method:`rs.freeze()`
+
+     - Makes the replica set member that ``mongosh`` is connected to 
+       ineligible to become primary for the specified duration.
```
jeff-allen-mongo, line 894: I know we are linking out to the manual for more info, but we could add a sentence here stating "The duration is specified in seconds." That way users have at least a baseline to run this method from this page.

jeff-allen-mongo (approved):
LGTM with a small thought.

aherlihy (approved):

PR #69, opened 2020-12-09
DOCSP-13313: [MONGOSH] Investigate if reconnecting MongoClient will interfere with sessions

steveren:
[Jira](https://jira.mongodb.org/browse/DOCSP-13313)
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-13313/reference/methods)

### DIFF
--- a/source/includes/admonitions/connection-resets.rst
+++ b/source/includes/admonitions/connection-resets.rst
@@ -1,8 +1,13 @@
 .. important::
 
    Methods which reset the shell's connection to the server trigger a
-   termination of all open :manual:`sessions </reference/server-sessions>`.
-   The following methods reset the shell's connection:
+   termination of all open server :manual:`sessions
+   </reference/server-sessions>`. When a session ends, all in-progress
+   operations are terminated as well, and the client must re-authenticate if 
+   authentication is enabled.
+
+   The following methods reset the shell's
+   connection:
 
    - ``db.auth``
    - ``Mongo.setReadPref``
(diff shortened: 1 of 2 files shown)


source/includes/admonitions/connection-resets.rst
```diff
@@ -0,0 +1,9 @@
+.. important::
+
+   Methods which reset the shell's connection to the server trigger a
+   termination of all open :manual:`sessions </reference/server-sessions>`.
```
jeff-allen-mongo, line 4: I think we should specify whether we mean client or server sessions.

jeff-allen-mongo, line 4: It may also be worth mentioning:

- Killing a session kills any in-progress operations in the session.
- Killing a session means the client will have to reauthenticate if auth is enabled.

jeff-allen-mongo (commented):
Thanks Steve! I have a few questions we may want to provide additional clarity on.

source/includes/admonitions/connection-resets.rst
steveren: Done.

steveren: Done.

steveren (commented):
Thanks @jeff-allen-mongo, this is updated with your feedback.

aherlihy (approved):

jeff-allen-mongo (approved):
LGTM

PR #72, opened 2020-12-22
DOCSP-13569: [MONGOSH] Mongosh fails to connect to node in state STARTUP2

steveren:
[Jira](https://jira.mongodb.org/browse/DOCSP-13569)
[Staged](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-13569/connect#connect-to-a-replica-set)

### DIFF
--- a/source/connect.txt
+++ b/source/connect.txt
@@ -162,6 +162,16 @@ To connect to a replica set:
 
         mongosh "mongodb://mongodb0.example.com.local:27017,mongodb1.example.com.local:27017,mongodb2.example.com.local:27017/?replicaSet=replA"
 
+.. note::
+
+   ``mongosh`` adds the ``directConnection=true`` query parameter to the
+   connection string automatically unless at least one of the following is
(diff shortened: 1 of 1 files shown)


andf-mongodb (approved):
LGTM, with the following observation:

• We don't document directConnection=true anywhere: neither in [the mongosh options page](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-13569/reference/options), nor even in [the old mongo shell options list](https://docs.mongodb.com/manual/reference/program/mongo/). It seems strange that we are telling users:

IMPORTANT: We now include a specific option as a new default unless you do one of three things. The option itself shouldn't concern you though ...

However, upstream tickets all seem quite happy with this approach, so I don't see anything for DOCS to do.

rose-m (approved):
Thanks for putting that in the docs! 😊

PR #80, opened 2021-01-06
DOCSP-13662: v0.7.0 updates

steveren:
Jira:
https://jira.mongodb.org/browse/DOCSP-13662
https://jira.mongodb.org/browse/DOCSP-13741
https://jira.mongodb.org/browse/DOCSP-13749
https://jira.mongodb.org/browse/DOCSP-13739

Staged:
https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-13662/

### DIFF
--- a/source/install.txt
+++ b/source/install.txt
@@ -101,6 +101,11 @@ Procedure
       
             .. include:: /includes/steps/install-shell-linux-generic.rst
 
+.. important:: Windows Users
+
+   ``mongosh`` preferences are stored by default in the directory
+   ``C:\Users\<username>'\.mongodb\mongosh``.
+
 Next Steps
 ----------
 
--- a/source/install.txt
+++ b/source/install.txt
@@ -39,7 +39,7 @@ Procedure
       .. note::
 
          ``mongosh`` preferences are stored by default in the directory
-         ``C:\Users\<username>'\.mongodb\mongosh``.
+         ``C:\Users\<username>\AppData\Roaming``.
 
    .. tab::
       :tabid: macOS
--- a/source/install.txt
+++ b/source/install.txt
@@ -36,6 +36,11 @@ Procedure
 
       .. include:: /includes/steps/install-shell-windows.rst
 
+      .. note::
+
+         ``mongosh`` preferences are stored by default in the directory
+         ``C:\Users\<username>'\.mongodb\mongosh``.
+
    .. tab::
       :tabid: macOS
 
--- a/source/reference/methods.txt
+++ b/source/reference/methods.txt
@@ -412,12 +412,17 @@ Cursor Methods
 
        .. arrayAccess
 
-   * - :method:`cursor.batchSize()`
+   * - ``cursor.batchSize()``
 
      - Controls the number of documents MongoDB will return to the
        client in a single network message.
 
-       .. clone
+       In the following example, the query returns results in batches of
+       100: 
+
+       .. code-block:: sh
+
+          db.myCollection.find().batchSize(100)
 
    * - :method:`cursor.close()`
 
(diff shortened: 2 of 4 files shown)


source/reference/methods.txt
```diff
@@ -412,12 +412,17 @@ Cursor Methods
...
        .. arrayAccess
 
-   * - :method:`cursor.batchSize()`
+   * - ``cursor.batchSize()``
 
      - Controls the number of documents MongoDB will return to the
        client in a single network message.
 
-       .. clone
+       In the following example, the query returns results in batches of
```
jeff-allen-mongo, line 420: [nit] Consider tightening this up to "The following example query returns..."

source/reference/methods.txt
```diff
@@ -709,6 +714,24 @@ Native Methods
...
+       .. code-block:: code
+
+          > print("hello world")
+          hello world
+
+          > x = "example text"
+          > print(x)
+          example text
+
+   * - :method:`sleep()`
```
jeff-allen-mongo, line 731: This links me to the top of the current page. I assume this isn't intentional?

source/run-commands.txt
```diff
@@ -47,6 +47,22 @@ the server.
    provides consistent behavior by always throwing an exception in these
    scenarios.
 
+Shell Commands
```
jeff-allen-mongo, line 50: I think "shell commands" is easy to conflate with "shell methods". What do you think about making the heading the specific task this command accomplishes, like "Clear ``mongosh`` Console" (or something like that), and explaining the ``cls`` method that way.

If we end up with more commands that would fit in here, we can consider a table structure at a later time.

source/install.txt
```diff
@@ -101,6 +101,11 @@ Procedure
       
             .. include:: /includes/steps/install-shell-linux-generic.rst
 
+.. important:: Windows Users
```
jeff-allen-mongo, line 104: I think we can move this so it only appears under the Windows tab.

source/install.txt
```diff
@@ -101,6 +101,11 @@ Procedure
...
+
+   ``mongosh`` preferences are stored by default in the directory
```
jeff-allen-mongo, line 106: As part of the same ticket (DOCSP-13741), there is also an update to where log files are stored. We may also need to update https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-13662/logs accordingly?

jeff-allen-mongo (commented):
Thanks Steve! Left a few comments / suggestions.

source/install.txt
jeff-allen-mongo: Do we need a parallel blurb for Linux / macos tabs which explain where to find Preferences on those operating systems as well?

steveren: Moved it under the windows tab and changed it to a note. Judging from the Jira ticket, this is a behavioral oddity which is only noteworthy for windows users.

source/reference/methods.txt
steveren: Done.

source/reference/methods.txt
steveren: Fixed.

source/run-commands.txt
steveren: Done.

source/install.txt
```diff
@@ -36,6 +36,11 @@ Procedure
 
       .. include:: /includes/steps/install-shell-windows.rst
 
+      .. note::
```
jeff-allen-mongo, line 39: Would it not be better to have this blurb appear under the macOS and Linux tabs, and have the Windows-specific blurb under the Windows tab?

jeff-allen-mongo (approved):
LGTM with one suggested fix.

source/install.txt
steveren: Fixed.

aherlihy (approved):

addaleax (approved):

PR #85, opened 2021-02-16
FLE documentation updates

jeff-allen-mongo:
These updates cover all tickets in the main [FLE ticket](https://jira.mongodb.org/browse/DOCSP-14006) (which has linked subtasks)

Note: We will be adding a new section for Azure / GCP support for FLE (with examples) shortly after this PR. The main purpose of this PR is to add the new FLE methods to the `mongosh` methods page.

Staging:

- [Install docs](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-13918/install/#procedure) (Homebrew notice)
- [mongosh methods](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-13918/reference/methods/)

### DIFF
--- a/source/reference/methods.txt
+++ b/source/reference/methods.txt
@@ -356,6 +356,62 @@ Connection Methods
 
      - Description
 
+   * - :method:`Mongo()`
+
+     - JavaScript constructor to instantiate a database connection from
+       the ``mongo`` shell or from a JavaScript file.
+
+       The :method:`Mongo()` method takes a parameter called
+       ``ClientSideFieldLevelEncryptionOptions``, with the following options:
+
+       .. list-table::
+          :widths: 15 20 40
+          :header-rows: 1
+
+          * - Option
+
+            - Type
+
+            - Description
+
+          * - ``keyVaultClient``
+
+            - :method:`Mongo()` connection object
+
+            - *(Optional)* The MongoDB cluster hosting the key vault
+              collection. Omit to use the current database connection
+              as the key vault host.
+
+          * - ``keyVaultNamespace``
+
+            - String
+
+            - *(Required)* The full namespace of the key vault collection.
+
+          * - ``kmsProvider``
+
+            - Document
+
+            - *(Required)* The Key Management Service (KMS) used by
+              client-side field level encryption for managing a Customer
+              Master Key (CMK).
+
+          * - ``schemaMap``
+
+            - Document
+
+            - *(Optional)* The automatic client-side field level encryption
+              rules specified using the JSON schema Draft 4 standard syntax
+              and encryption-specific keywords.
+
+          * - ``bypassAutoEncryption``
+
+            - Boolean
+
+            - *(Optional)* Specify ``true`` to bypass automatic client-side
+              field level encryption rules and perform explicit (manual)
+              per-field encryption.
+
    * - :method:`Mongo.getDB()`
 
      - Returns a database object.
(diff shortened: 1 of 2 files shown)


source/reference/methods.txt
```diff
@@ -356,6 +356,62 @@ Connection Methods
 
      - Description
 
+   * - :method:`Mongo()`
+
+     - JavaScript constructor to instantiate a database connection from
+       the ``mongo`` shell or from a JavaScript file.
+
+       The :method:`Mongo()` method takes a parameter called
```
addaleax, line 364: Note that, like in the old shell, it takes two arguments -- the first one is still the connection string

source/reference/methods.txt
```diff
@@ -356,6 +356,62 @@ Connection Methods
...
+
+            - Description
+
+          * - ``keyVaultClient``
+
+            - :method:`Mongo()` connection object
+
+            - *(Optional)* The MongoDB cluster hosting the key vault
+              collection. Omit to use the current database connection
+              as the key vault host.
```
addaleax, line 383: I think this is misleading even in the old shell's documentation. It's not the current database connection (i.e. the one referred to by `db`), it's the newly created `Mongo` object's own one (i.e. the one pointed to by the connection string)

source/reference/methods.txt
```diff
@@ -356,6 +356,62 @@ Connection Methods
...
+
+          * - ``keyVaultNamespace``
+
+            - String
+
+            - *(Required)* The full namespace of the key vault collection.
+
+          * - ``kmsProvider``
```
addaleax, line 391: Ouch -- I just noticed that this is actually a typo in the old shell's documentation. This should be `kmsProviders`, not `kmsProvider`. Unfortunately, we relied on the documentation here, so currently mongosh would expect `kmsProvider` -- I'll try to get this fixed within the next hour or so, so we're not releasing the wrong thing: https://jira.mongodb.org/browse/MONGOSH-604

source/reference/methods.txt
```diff
@@ -691,6 +747,88 @@ standalone and replica set deployments.
...
 
+Client-Side Field Level Encryption Methods
+------------------------------------------
+
+.. note:: Limitations
+
+   Automatic encryption is only available when ``mongosh`` is connected
+   to an Atlas cluster or a MongoDB Enterprise Server. For details, see
+   :manual:`Automatic Client-Side Field Level Encryption
+   </core/security-automatic-client-side-encryption/>`.
```
addaleax, line 758: It might be worth noting that all the methods here still work for non-enterprise servers, because they only perform manual encryption/decryption.

source/reference/methods.txt
jeff-allen-mongo: Ah sorry about that!

source/reference/methods.txt
jeff-allen-mongo: Good point. I could see that being a point of confusion for readers.

source/reference/methods.txt
jeff-allen-mongo: I've updated to defer to the old shell docs here and keep this description high level. I've created https://jira.mongodb.org/browse/DOCSP-14537 to address the confusion in the old shell docs, which we will get to in this sprint.

source/reference/methods.txt
jeff-allen-mongo: Updated.

jeff-allen-mongo:
Thank you for the comments @addaleax! I've updated this PR.

source/reference/methods.txt
```diff
@@ -356,6 +356,56 @@ Connection Methods
...
+             - Description
+       
+           * - ``host``
+       
+             - string
+       
+             - *Optional*
+             
+               The host, either in the form of ``<host>`` or 
+               ``<host><:port>``.
```
addaleax, line 383: Fwiw, this can be any valid mongodb connection string, and include e.g. database name as well. I think this applies to the old shell as well.

addaleax (approved):

PR #88, opened 2021-02-22
DOCSP-14691 add shell csfle section

andf-mongodb:
This PR adds a new CSFLE landing page, including:
-  Brief CSFLE overview, linking back to Manual
-  New tab layout for all four supported KMS providers:

    1. AWS KMS (previously supported, but undocumented)
    2. Azure Vault (new with 0.8.0)
    3. GCP KMS (new  with 0.8.0)
    4. Local Keystore (previously supported, but undocumented)

 _NOTES_ 
- I don't presently have a way of testing against an active KMS for any of these platforms, so I have provided my best guesses. Please validate if you, the reviewer, have such access!
- Please let me know if I can provide better, more practical example key names / location names / etc. Or if user params should be indicated in a different manner, perhaps with "< exampleText >". I hesitate to use caps or camelcase, supposing that upstream might not support case identically across all three cloud providers.
- This update only adds "Create a Key" instructions (+validate key, at the end). Let me know if we need to add more tasks, such as dealing with AltNames, or removing keys.
- I left off all optional params to these commands, like specifying KeyAltNames, or overriding cloud-provider-specific KMS URLs/ports. Let me know if this is required.

[ [STAGING LINK](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-14691-add-shell-fle-section/field-level-encryption/#client-side-field-level-encryption) ]

Thank you!

### DIFF
--- a/source/includes/steps-local-keyfile.yaml
+++ b/source/includes/steps-local-keyfile.yaml
@@ -1,0 +1,101 @@
+---
+title: "Launch the ``mongosh`` Shell."
+ref: launch-mongosh-shell
+level: 4
+content: |
+
+  Create a ``mongosh`` session without connecting to a running database
+  by using the :option:`--nodb` option:
+
+  .. code-block:: javascript
+
+     mongosh --nodb
+
+---
+title: "Generate an Encryption Key."
+ref: generate-local-key
+level: 4
+content: |
+
+  To configure client-side field level encryption for a locally managed
+  key, you must specify a base64-encoded 96-byte string with no line
+  breaks. Run the following command in ``mongosh`` to generate a key
+  matching these requirements:
+
+  .. code-block:: javascript
+
+     crypto.randomBytes(96).toString('base64')
+
+  You will need this key in the next step.
+
+---
+title: "Create the Encryption Configuration."
+ref: create-encryption-configuration
+level: 4
+content: |
+
+  In ``mongosh``, create a new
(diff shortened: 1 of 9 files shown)


source/includes/steps-local-keyfile.yaml
```diff
@@ -0,0 +1,101 @@
...
+content: |
+
+  To configure client-side field level encryption for a locally managed
+  key, you must specify a base64-encoded 96-byte string with no line
+  breaks. Run the following command on the command line (*not* within
+  ``mongosh``) to generate a key matching these requirements:
+
+  .. code-block:: javascript
+
+     echo $(head -c 96 /dev/urandom | base64 | tr -d '\n')
```
addaleax, line 14: It might be easier to tell people to run `crypto.randomBytes(96).toString('base64')` *inside* the shell, mongosh supports that (the old shell wouldn't) and it's cross-platform

addaleax (approved):

rose-m (approved):
Just a nitty typo :) Sorry for taking so long - had some technical issues 😅 I did try it with the `local` KMS, followed your example and worked. Also verified again with the documented API for all the options and couldn't find any issues 👍 
Thanks for writing all that up!

source/includes/steps-local-keyfile.yaml
andf-mongodb: Thanks! This is a great tip!

PR #89, opened 2021-02-23
DOCSP-14365 info for installing mongocryptd on linux

kanchana-mongodb:
- [JIRA](https://jira.mongodb.org/browse/DOCSP-14365)
- [Stage](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-14365/install/#procedure)

### DIFF
--- a/source/install.txt
+++ b/source/install.txt
@@ -106,11 +106,12 @@ Procedure
 
             .. note:: 
 
-               You will need Ubuntu 16.04 or higher version to install 
-               the ``mongocryptd`` binary in the |mdb-shell| ``.deb`` 
-               package. On older Ubuntu versions, you must 
+               The ``mongocryptd`` binary in the |mdb-shell| ``.deb`` 
+               package does not work on Ubuntu 16.04 or older. On 
+               Ubuntu 16.04 and older systems, you must 
                :manual:`install </reference/security-client-side-encryption-appendix/#installation>` 
-               the ``mongocryptd`` binary separately.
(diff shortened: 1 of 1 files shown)


source/install.txt
```diff
@@ -104,6 +104,14 @@ Procedure
          .. tab:: .deb
             :tabid: deb
 
+            .. note:: 
+
+               You will need Ubuntu 16.04 or higher version to install 
+               the ``mongocryptd`` binary in the |mdb-shell| ``.deb`` 
+               package. On older Ubuntu versions, you must 
+               :manual:`install </reference/security-client-side-encryption-appendix/#installation>` 
+               the ``mongocryptd`` binary separately.
```
jeff-allen-mongo, line 113: I think this is a little misleading. As I understand it, for Ubuntu versions higher than 16.04, they can just use the `mongocryptd` packaged with the shell. They shouldn't need to install anything.

I'd consider rewording to something like:

```
The :ref:`mongocryptd` packaged with the |mdb-shell| does not work on
Ubuntu 16.04 or older. On Ubuntu 16.04 and older systems, you must
:manual:`install </reference/security-client-side-encryption-appendix/#installation>` 
the ``mongocryptd`` binary separately to use its functionality in the |mdb-shell|.
```

jeff-allen-mongo (approved):
LGTM!

rose-m (approved):
From what I know about that issue that's basically exactly it 👍

PR #93, opened 2021-03-10
(DOCSP-15026): Load method

jeff-allen-mongo:
Staged: [Native Methods](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-15026/reference/methods/#native-methods)

### DIFF
--- a/source/reference/methods.txt
+++ b/source/reference/methods.txt
@@ -842,10 +842,10 @@ Native Methods
      - Loads and runs a JavaScript file in the shell.
 
        In ``mongosh``, scripts loaded with the ``load()`` method support
-       the ``__filename`` and ``__dirname`` Node.js environment
-       variables. These variables return the file name and directory of
-       the loaded script, respectively. The returned values are always
-       absolute paths.
+       the ``__filename`` and ``__dirname`` Node.js variables. These
+       variables return the file name and directory of the loaded
+       script, respectively. The returned values are always absolute
+       paths.
 
        In the legacy shell, you cannot not access the script's file name
        or directory using the ``load()`` method.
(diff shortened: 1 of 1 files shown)


source/reference/methods.txt
```diff
@@ -837,6 +837,18 @@ Native Methods
 
      - Description
 
+   * - :manual:`load() </reference/method/load/>`
+
+     - Loads and runs a JavaScript file in the shell.
+
+       In ``mongosh``, scripts loaded with the :method:`load()` method
+       support the ``__filename`` and ``__dirname`` Node.js objects.
```
addaleax, line 845: It might be a bit confusing to refer to these as objects, because they are strings and in JS terminology objects are a distinct type from strings? But I guess anyone familiar with Node.js would be familiar with that as well.

What might be worth pointing out is that these are always absolute paths.

addaleax (approved):

source/reference/methods.txt
jeff-allen-mongo: Thank you for the comments! I called them "objects" initially because I saw them listed here, [as nodejs Global Objects](https://www.tutorialspoint.com/nodejs/nodejs_global_objects.htm). However, it seems like "environment variables" might be more appropriate.

addaleax: Maybe we could call them `per-script variables`? That might still not be ideal, but given how `environment variables` also has a very specific meaning, I think calling them something like that might be better.

jeff-allen-mongo: Ah, I'd gleaned from [this tutorial](https://www.digitalocean.com/community/tutorials/nodejs-how-to-use__dirname) that they might be environment variables, but happy to use different terminology if "environment variables" is not appropriate.

I'm inclined to just generalize to say  "__filename and __dirname variables", since I worry that the "per-script" terminology might be a little confusing.

PR #99, opened 2021-04-27
(DOCSP-15123) Add note to tlsCertificateSelector option

atsansone:


addaleax (approved):
I don't know if it's possible from a legal point of view, but if you want you can add a reference to the screenshot mentioned in the ticket, so that it's a bit more obvious what to look for :)

PR #103, opened 2021-05-07
DOCSP-15697 Create an API for global shell settings

davemungo:


source/reference/configure-shell-settings.txt
```diff
@@ -0,0 +1,166 @@
...
+     - Enables sending annonymized tracking and diagnostic data to
+       MongoDB. 
+
+   * - ``historyLength``
+     - integer
+     - 1000
+     - The number of items to store in ``mongosh`` REPL's history file.
+
+   * - ``inspectDepth``
+     - integer or "infinity"
```
davemungo, line 74: Not sure about this one, I have two questions. 
1. It seems that almost any string will set infinity (other than something like "3" which sets integer 3). Is this expected? 
2. Depth of inspection works as expected, but the .pretty() modifier doesn't work consistently when inspectDepth is changed. Is this expected?

See   https://gist.github.com/davemungo/1b44d6f6316edc07eb3df9a858160420 for various examples.

addaleax: Regarding 1.: I don’t think it’s expected that you plug in strings where a numeric value is expected – we could probably provide a better error message in that case, though.

Regarding 2.: This is a bit related to https://jira.mongodb.org/browse/MONGOSH-684, but currently, `.pretty()` is a no-op in mongosh.

davemungo: @addaleax  I think I added this just after you started the rest of the review. Would you take a look? Thanks

source/reference/configure-shell-settings.txt
```diff
@@ -0,0 +1,166 @@
...
+
+   * - ``batchSize``
+     - integer
+     - 20
+     - The number of items displayed per cursor iteration
+
+   * - ``enableTelemetry``
+     - boolean
+     - false
+     - Enables sending annonymized tracking and diagnostic data to
```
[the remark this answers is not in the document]
davemungo: Done

source/reference/configure-shell-settings.txt
```diff
@@ -0,0 +1,166 @@
...
+update the ``batchSize`` to limit the number of items returned by a
+cursor.
+
+.. code-block:: javascript
+
+   config.set("batchSize", 3) 
+
+
+Future ``db.collection.find()`` operations will only return 3 documents
+per cursor itteration.
```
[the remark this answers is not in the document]
davemungo: Done

source/reference/configure-shell-settings.txt
```diff
@@ -0,0 +1,166 @@
...
+
+.. important::
+
+   You must use different quotation marks for the :option:` --eval`
+   expression and the ``config`` property. That is, single quotes for
+   one and double quotes for the other. 
+
+.. code-block::
+
+   mongosh --port 27500 --eval '{{config.set("enableTelemetry", true)}}'
```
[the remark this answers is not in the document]
davemungo: Done

source/reference/configure-shell-settings.txt
```diff
@@ -0,0 +1,166 @@
...
+   You must use different quotation marks for the :option:` --eval`
+   expression and the ``config`` property. That is, single quotes for
+   one and double quotes for the other. 
+
+.. code-block::
+
+   mongosh --port 27500 --eval '{{config.set("enableTelemetry", true)}}' 
+
+``mongosh`` returns additional information along with the result of the
+API call. In this case MongoDB is running on port 27500.
```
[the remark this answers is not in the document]
davemungo: Done

source/reference/configure-shell-settings.txt
```diff
@@ -0,0 +1,166 @@
...
+     - Description
+
+   * - ``batchSize``
+     - integer
+     - 20
+     - The number of items displayed per cursor iteration
+
+   * - ``enableTelemetry``
+     - boolean
+     - false
```
addaleax, line 64: This is currently an opt-out property, so maybe this should be `true`?

davemungo: Done 

I may have changed it when I installed the shell initially

source/reference/configure-shell-settings.txt
davemungo: I think it might be a common mistake to enter "3" instead of 3, although it doesn't seem to matter since the "3" gets converted to a 3 anyway. 

I'll update it to use the js 'Infinity' (no quotes). 

But it seems like a validation error that a string is accepted and will be converted to mean 'Infinity' (unless it can be converted to int). Do you want me to open a ticket for it?

davemungo: https://jira.mongodb.org/browse/MONGOSH-758

addaleax (approved):

PR #107, opened 2021-05-12
(DOCSP-15892): Read preference difference between mongosh and mongo

jeff-allen-mongo:
JIRA: [DOCSP-14221](https://jira.mongodb.org/browse/DOCSP-14221)

Staged: [Compatibility Changes](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-14221/reference/compatibility/)

### DIFF
--- a/source/reference/compatibility.txt
+++ b/source/reference/compatibility.txt
@@ -52,3 +52,22 @@ the methods listed in the :guilabel:`Alternative Methods` column.
        - :method:`db.collection.findOneAndUpdate()`
        - :method:`db.collection.bulkWrite()`
 
+Read Preference Behavior
+------------------------
+
+When using ``mongosh`` to connect directly to a :term:`secondary`
+replica set member, if you specify a default read preference, that
+preference is honored for commands run against that member. For example,
+specifying ``readPreference=primary`` directs read operations to the
+replica set :term:`primary`. 
+
+To set a default read preference, you can either:
+
+- Use the :urioption:`readPreference` connection string option when
+  connecting to the node.
+
+- Use the :method:`Mongo.setReadPref()` method.
+
+In the legacy ``mongo`` shell, the default read preference is *not*
+honored when connected to a :term:`secondary` replica set member. Read
+operations are always directed to the member to which you are connected.
--- a/source/reference/compatibility.txt
+++ b/source/reference/compatibility.txt
@@ -76,7 +76,7 @@ To specify a read preference, you can use either:
 
 - The :method:`Mongo.setReadPref()` method.
 
-``rs.secondaryOk()`` was required in the legacy ``mongo`` shell to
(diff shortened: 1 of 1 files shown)


source/reference/compatibility.txt
```diff
@@ -52,3 +52,22 @@ the methods listed in the :guilabel:`Alternative Methods` column.
...
        - :method:`db.collection.bulkWrite()`
 
+Read Preference Behavior
+------------------------
+
+When using ``mongosh`` to connect directly to a :term:`secondary`
+replica set member, if you specify a default read preference, that
+preference is honored for commands run against that member. For example,
+specifying ``readPreference=primary`` directs read operations to the
+replica set :term:`primary`.
```
addaleax, line 62: Are you sure about this? I have a replset at ports 27017, 27018, 27019, with 27107 being the primary, but I get this behavior:

```js
$ ./bin/mongosh 'mongodb://localhost:27018/?readPreference=primary'
Current Mongosh Log ID: 609c5499b477eb0dfc6f6fd0
Connecting to:          mongodb://localhost:27018/?readPreference=primary&directConnection=true&serverSelectionTimeoutMS=2000
[...]
replset [direct: secondary]> db.test.find()
MongoError: not master and sl***Ok=false
```

Which matches my understanding of what happens -- when mongosh connects to a single host, it adds `directConnection=true` to the connection string, which is documented to require enabling *secondary* reads in order to be able to read: https://docs.mongodb.com/drivers/node/fundamentals/connection/#direct-connection

I.e. the shell still only maintains a connection to the secondary, and the secondary does not forward the read request to the primary, as far as I can tell, and one would need to set the read pref to `secondary`, `secondaryPreferred` or `primaryPreferred` in order to be able to read (although there would be no difference between those).

addaleax (commented):
By the way, this might also be a good place to mention that `show databases`/`show dbs` and `show collections` always use a readPreference of `primaryPreferred`.

source/reference/compatibility.txt
jeff-allen-mongo: 🤦‍♂️ I see what I did wrong. I was trying to test using the same config you've given in your example, but had accidentally specified `readPreference=secondary` instead of `primary`.

Can you please verify how this differs from the legacy shell? In the legacy shell, when I connect directly to a secondary with `?readPreference=primary`, I receive the same error / behavior.

addaleax: I think the only difference is that the legacy shell additionally needed a `mongo.setSecondaryOk()` call in order to be able to read from secondaries, which in mongosh is unsupported (i.e. the read preference takes care of everything).

jeff-allen-mongo: Thank you @addaleax ! I've updated this section and also added a new section for the `show` helpers mentioned in the ticket.

addaleax (approved):

PR #111, opened 2021-05-19
(DOCSP-16174): install page fixups

jeff-allen-mongo:
JIRA: [DOCSP-16174](https://jira.mongodb.org/browse/DOCSP-16174)
Staged: [install mongosh](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-16174/install/)

Summary of changes:

- On macOS, we’re using .zip instead of .tgz, so the filenames and extraction commands need to be adjusted
- The file suggests copying mongosh to e.g. /usr/local/bin – The mongosh-mongocryptd from the package needs to be copied as well, to /usr/local/libexec or /usr/local/bin as well (or generally ../libexec or ./ when seen relative from the mongosh binary itself)

Note: The JIRA ticket also mentions that the “macOS may prevent mongosh from running after installation” seems outdated, but I could not verify this in testing, so leaving this step in the procedure for now.

### DIFF
--- a/source/includes/steps-install-shell-macos.yaml
+++ b/source/includes/steps-install-shell-macos.yaml
@@ -9,10 +9,30 @@ source:
   ref: download-archive
 ref: download-archive-macos
 ---
-source:
-  file: steps-install-shell-base.yaml
-  ref: extract-archive
-ref: extract-archive-macos
+title: "Extract the files from the downloaded archive."
+ref: extract-zip-macos
+level: 4
+content: |
+
+   Run the following command from the directory containing the
+   ``mongosh`` ``.zip`` archive:
+   
+   .. code-block:: sh
+
+      unzip mongosh-{version}-darwin-x64.zip
+
+   The extracted ``bin`` folder contains two binaries: ``mongosh`` and
+   ``mongocryptd-mongosh``.
+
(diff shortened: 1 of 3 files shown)


source/includes/steps-install-shell-macos.yaml
```diff
@@ -9,10 +9,26 @@ source:
...
+      unzip mongosh-{version}-darwin-x64.zip
+
+   If your web browser automatically extracts the archive as
+   part of the download or you extract the archive without the
+   ``unzip`` command, you may need to run the following command to
+   make ``mongosh`` executable:
+
+   .. code-block:: sh
+
+      chmod +x /bin/mongosh
```
addaleax, line 31:
```suggestion
      chmod +x bin/mongosh
```

Unless this would actually be installed in /bin, which would be fairly untypical?

addaleax (approved):

source/includes/steps-install-shell-macos.yaml
jeff-allen-mongo: Updated and tweaked this a bit to clarify where you need to run this command. Thanks for flagging this.

PR #115, opened 2021-06-02
DOCSP-15687 numeric values stored as int32

davemungo:
This change updates the type information to state that certain numbers are entered in mongosh as int32 rather than double. It also adds a warning that updating the same database with mongo and mongosh may result in data inconsistency.

### JIRA
https://jira.mongodb.org/browse/DOCSP-15687

### STAGING
https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-15687-numeric-values-stored-as-int32/reference/data-types/

### DIFF
--- a/source/reference/data-types.txt
+++ b/source/reference/data-types.txt
@@ -130,9 +130,10 @@ operation in ``mongosh``:
 NumberLong
 ~~~~~~~~~~
 
-``mongosh`` treats all numbers as floating-point values by default.
-``mongosh`` provides the ``NumberLong()`` wrapper to handle 64-bit
-integers.
+If a number can be converted to a 32-bit ``int``, ``mongosh`` will
+store it in that form. If not, ``mongosh`` defaults to storing the
+number as a ``double``. The ``NumberLong()`` wrapper can convert the
+type to a 64-bit integer.
 
 The ``NumberLong()`` wrapper accepts the long as a string:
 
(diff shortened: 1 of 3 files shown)


mmarcon:
That sentence looks good to me.

Aside from that part specifically, I'd suggest moving away from the number types that were present in the legacy shell, namely `NumberLong`, `NumberInt`, and `NumberDecimal` and encourage users to use the driver-compatible types instead: `Long`, `Int32`, and `Decimal128` respectively.

I guess is fair to have a section where we mention that the legacy types are still supported but the new, driver-compatible ones should be preferred.

I don't know if you had a separate ticket for that but I thought I'd mention it here.

source/reference/data-types.txt
```diff
@@ -285,13 +287,13 @@ returned:
...
-       | **{ "_id": 4, "val": NumberLong(10), "description": "Long" }**
+     - | **{ "_id": 3, "val": 10, "description": "Int" }**
+       | **{ "_id": 4, "val": 10, "description": "Long" }**
        | **{ "_id": 5, "val": NumberDecimal( "10.0" ), "description": "Decimal" }**
 
    * - **{ val: NumberDecimal( "10" ) }**
-     - | **{ "_id": 3, "val": 10, "description": "Double" }**
-       | **{ "_id": 4, "val": NumberLong(10), "description": "Long" }**
+     - | **{ "_id": 3, "val": 10, "description": "Int" }**
+       | **{ "_id": 4, "val": 10, "description": "Long" }**
```
addaleax, line 296: This was surprising, thanks for making us notice :) I’ve opened https://jira.mongodb.org/browse/MONGOSH-809 and we’ve decided that we’ll change this so that the output here should be `Long(10)` instead of just `10`.

addaleax (approved):

source/reference/data-types.txt
davemungo: Cool! Glad to help. 

Should I wait for the fix to go in or go with the current output and update later?

davemungo:
@mmarcon  Created https://jira.mongodb.org/browse/DOCSP-16365 to update the type casting methods

source/reference/data-types.txt
```diff
@@ -322,8 +324,8 @@ string alias ``"decimal"`` or ``19``, the numeric code for the
 
 .. _check-types-in-shell:
       
-Check Types in the ``mongo`` Shell
-----------------------------------
+Check Types in the MongoDB shell, ``mongosh``
```
jeff-allen-mongo, line 327: Sorry I missed this in the review. Can you please make this "MongoDB Shell" ? (title case).

davemungo: Done.

jeff-allen-mongo (approved):

PR #116, opened 2021-06-07
(DOCSP-16326): Remove mentions of .load

jeff-allen-mongo:
This update removes mention of `.load` and replaces them with the `load()` method.

JIRA: [DOCSP-16326](https://jira.mongodb.org/browse/DOCSP-16326)

STAGED: [Execute a JavaScript File](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-16326/write-scripts/#execute-a-javascript-file)

### DIFF
--- a/source/changelog.txt
+++ b/source/changelog.txt
@@ -134,7 +134,7 @@ v0.9.0
 
 New features in this release:
 
-- Support for the :manual:`load() </reference/method/load/>` method.
+- Support for the :method:`load()` method.
 
 - Support for AWS IAM authentication.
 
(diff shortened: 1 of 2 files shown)


mmarcon:
LGTM. I don't know if this is the right place, of if the `load()` documentation is the right place, or if it should be in both places but somewhere it's worth mentioning that `mongosh` exposes a variable `__dirname` that can be used to load files from the same directory of another loaded file without having to specify absolute paths. E.g.

```javascript
load(__dirname + '/prompt.js');
load(__dirname + '/sql.js');
load(__dirname + '/explorer.js');
```

jeff-allen-mongo:
Thanks for the comment @mmarcon . I created https://jira.mongodb.org/browse/DOCSP-16464 to track the __dirname update.

PR #121, opened 2021-06-25
 DOCSP-16365 align mongosh types with driver types

davemungo:
This change updates the information on type changes in mongosh compared to mongo shell. 

In addition to the content changes, the layout has been reorganized and several examples have been consolidated in one section.

STAGING:
https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-16365-align-mongosh-types-with-driver-types/reference/data-types/

### DIFF
--- a/source/reference/compatibility.txt
+++ b/source/reference/compatibility.txt
@@ -115,8 +115,14 @@ Numeric Values
 --------------
 
 The legacy ``mongo`` shell stored numerical values as ``doubles`` by
-default. In ``mongosh`` numbers are stored as 32 bit ``int`` types when
-possible or else as ``double``. 
+default. In ``mongosh`` numbers are stored as 32 bit integers,
+``Int32``, or else as ``Double`` if the value cannot be stored as an
+``Int32``. 
+
+MongoDB Shell continues to support the numeric types that are supported
+in ``mongo`` shell. However, the preferred types have been updated to
+better align with the MongoDB :ecosystem:`drivers </drivers>`. See
+:ref:`mongosh Data Types <mongo-shell-data-type>` for more information.
 
 The preferred types for numeric variables are different in MongoDB 
 Shell than the types suggested in the legacy ``mongo`` shell. The types
@@ -141,47 +147,6 @@ in ``mongosh`` better align with the types used by the MongoDB Drivers.
 
    Data types may be stored inconsistently if you connect to the same
    collection using both ``mongosh`` and the legacy ``mongo`` shell. 
-
-Consider the ``typeExample`` collection. This collection consists of
-two identical documents, ``{ "a": 1 }``. The first document was created
-in the legacy ``mongo`` shell, the second document was created in
-``mongosh``. 
-
-We can use the :expression:`$type` operator in an aggregation pipeline
-to see the type that was assigned in each shell. 
-
-.. code-block:: javascript
-
-   db.typeExample.aggregate( 
-     [ 
-       {
-          $project: 
-            { 
-              "valueType": 
-                {
-                  "$type": "$a"
-                },
-              "_id": 0
-            }
-       }
-     ]
-   )
-
-In the first document, created in the legacy ``mongo`` shell, the value
-was stored as a ``double``. In the ``mongosh`` document the value was
-stored as type ``int``.
-
-.. code-block:: javascript
-   .. :copyable: false
-
-   [
-     { 
-        valueType: 'double'  // inserted in legacy mongo shell
-     },
-     {
-        valueType: 'int'     // inserted in mongosh
-     }
-   ]
 
 .. seealso::
 
--- a/source/reference/data-types.txt
+++ b/source/reference/data-types.txt
@@ -1,6 +1,6 @@
-=========================
-Data Types in ``mongosh``
-=========================
+======================
+``mongosh`` Data Types
+======================
 
 .. default-domain:: mongodb
 
@@ -10,22 +10,24 @@ Data Types in ``mongosh``
    :depth: 2
    :class: singlecol
 
-MongoDB :term:`BSON` provides support for additional data types than
-:term:`JSON`. :ecosystem:`Drivers </drivers>` provide native support for
-these data types in host languages and ``mongosh`` also provides several
-helper classes to support the use of these data types. See the
-:manual:`Extended JSON </reference/mongodb-extended-json>` reference for
-additional information.
-
 .. _mongo-shell-data-type:
 
-Types
------
+MongoDB Server stores data using the
+:manual:`BSON </reference/bson-types>` format which supports some
+additional data types that are not available using the :term:`JSON`
+format. Compared to the legacy ``mongo`` shell, MongoDB Shell
+(``mongosh``) has type handling which is better aligned with the
+default types used by the MongoDB :ecosystem:`drivers </drivers>`. 
+
+This document highlights changes in type usage between ``mongosh`` and
+the legacy ``mongo`` shell. See the 
+:manual:`Extended JSON </reference/mongodb-extended-json>` reference for
+additional information on supported types.
 
 .. _mongo-shell-date-type:
 
 Date
-~~~~
+----
 
 ``mongosh`` provides various methods to return the date, either as a
 string or as a ``Date`` object:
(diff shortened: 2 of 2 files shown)


source/reference/compatibility.txt
```diff
@@ -115,8 +115,14 @@ Numeric Values
 --------------
 
 The legacy ``mongo`` shell stored numerical values as ``doubles`` by
-default. In ``mongosh`` numbers are stored as 32 bit ``int`` types when
-possible or else as ``double``. 
+default. In ``mongosh`` numbers are stored as 32 bit integers,
+``Int32``, or else as ``Double`` if the value cannot be converted to an
```
addaleax, line 119:
```suggestion
``Int32``, or else as ``Double`` if the value cannot be stored as an
```

(because there is no real conversion taking place here)

source/reference/data-types.txt
```diff
@@ -38,325 +40,495 @@ string or as a ``Date`` object:
...
-well.
-
-To print the value of the variable, type the variable name in the
-shell, as in the following:
+.. note::
 
-.. code-block:: javascript
+   In the legacy :binary:`~bin.mongo` shell ``NumberLong()`` accepted
+   either a string or integer value. In ``mongosh``, ``NumberLong()``
+   only accepts string values. ``Long()`` accepts integers and strings.
```
addaleax, line 111:
```suggestion
   only accepts string values. ``Long()`` accepts strings as well.
```

`Long` doesn’t really accept single arbitrary integer values, it accepts *two* integer values representing the low/high bits of the number. Users should always pass strings, practically speaking.

source/reference/data-types.txt
```diff
@@ -38,325 +40,495 @@ string or as a ``Date`` object:
...
+Equality and Sort Order
+~~~~~~~~~~~~~~~~~~~~~~~
 
-   new ObjectId
+Values of the ``Decimal128`` type are compared and sorted with other
+numeric types based on their actual numeric value.  Numeric values
+of the binary-based ``Double`` type generally have approximate
+representations of decimal-based values and may not be exactly
+equal to their decimal representations. You can use the
+``Decimal128()`` constructor to check the equality of decimal values.
```
addaleax, line 154: Dumb question – how?

source/reference/data-types.txt
```diff
@@ -38,325 +40,495 @@ string or as a ``Date`` object:
...
 
-   :method:`ObjectId`
+.. _check-types-in-shell:
+      
+Type Checking
+-------------
 
-.. _shell-type-long:
+To determine the type of fields, ``mongosh`` provides ``instanceof``
+and ``typeof`` operators as well as the :query:`$type` query operator.
```
addaleax, line 163: `instanceof` does not work reliably in the shell – For example, BSON values coming from a server response and BSON values created by the user won’t be considered `instanceof` the same base class

(It’s also a best practice to avoid `instanceof` as much as possible in JS in general, I would say)

source/reference/data-types.txt
```diff
@@ -38,325 +40,495 @@ string or as a ``Date`` object:
...
-   NumberDecimal("1000.55")
-   
-The value is stored in the database as follows:
-
-.. code-block:: javascript
+   ISODate("2012-12-19T06:01:17.171Z")
 
-   NumberDecimal("1000.55")
+To verify the type, use the ``instanceof`` operator, as in the
+following:
```
addaleax, line 239: Ditto here – if you really **need** to verify that something is a Date, you can use `Object.prototype.toString.call(value) === [object Date]`

source/reference/data-types.txt
davemungo: Not sure what you mean, Long() currently accepts a single integer value: 

db.types.insertOne({ "a": 1 })
db.types.insertOne({ "a": Long(1) })
db.types.find({"a": {$type: "long"} })
[ { _id: ObjectId("60d648ed5b37cfdf17351b5b"), a: Long("1") } ]

Maybe it shouldn't do that, but it appears to work on regular integers.

source/reference/compatibility.txt
davemungo: Done

source/reference/data-types.txt
addaleax: `Long(1)` is short for `Long(1, 0)`, yes.

People should still either always pass in strings or two integers, not one.

davemungo: Done.

source/reference/data-types.txt
davemungo: The line was a summary of the prior copy. I tried making it work using some hokey wrapping, but it's probably better just to drop the statement.

source/reference/data-types.txt
davemungo: Done.

source/reference/data-types.txt
addaleax: Yeah, I think it might be better to either tell users how to check for equality (I guess comparing the `.toString()` results?) or not mention it 👍

source/reference/data-types.txt
davemungo: Done

davemungo:
I've removed the instanceof and typeof examples and made the other suggested updates. Would you have a look? Thanks!

addaleax (approved):

source/reference/data-types.txt
```diff
@@ -38,325 +40,451 @@ string or as a ``Date`` object:
...
 
-You can use the ``new`` operator with the ``ISODate()`` constructor as
-well.
+      db.types.insertOne(
+                          {
+                              "_id": 3,
+                              "value": Long("1"),
+                              "expectedType": "Long" 
+                          }
+                        )
```
jeff-allen-mongo, line 105: The indentation here is a little strange, no? Should be something like:

```suggestion
      db.types.insertOne(
         {
            "_id": 3,
             "value": Long("1"),
             "expectedType": "Long" 
         }
      )
```

source/reference/data-types.txt
```diff
@@ -38,325 +40,451 @@ string or as a ``Date`` object:
...
+The ``Decimal128`` :manual:`BSON type </reference/bson-types>`
+uses the IEEE 754 decimal128 floating-point numbering format which
+supports 34 decimal digits (i.e. significant digits) and an exponent
+range of −6143 to +6144.
 
 .. code-block:: javascript
 
-   myDate instanceof Date
-   myDateInitUsingISODateWrapper instanceof Date
+      db.types.insertOne(
```
jeff-allen-mongo, line 134: Same comment about indentation here.

source/reference/data-types.txt
```diff
@@ -38,325 +40,451 @@ string or as a ``Date`` object:
...
 
-To verify the type, use the ``typeof`` operator, as in the following:
+The :node-api:`Int32() <Int32.html>` constructor can be used to
+explicitly specify 32-bit integers.
 
 .. code-block:: javascript
 
-   typeof myDateString
+      db.types.insertOne(
+                          {
```
jeff-allen-mongo, line 76: Same comment about indentation here.

source/reference/data-types.txt
```diff
@@ -38,325 +40,451 @@ string or as a ``Date`` object:
...
+                        )
 
-To print the value of the variable, type the variable name in the
-shell, as in the following:
+.. note::
 
-.. code-block:: javascript
+   In the legacy :binary:`~bin.mongo` shell ``NumberLong()`` accepted
+   either a string or integer value. In ``mongosh``, ``NumberLong()``
+   only accepts string values. :node-api:`Long() <Long.html>` provides
```
jeff-allen-mongo, line 111: I'm a little confused by the inclusion of this last sentence about `Long`. It seems to come a little out of the blue since we were just talking about a different type (`NumberLong()`). Maybe I'm misunderstanding something here though, so maybe you can just clarify in a comment if you and @addaleax feel this is clear enough as written.

addaleax: `Long(string)` is the preferred version to create an int64 value – `NumberLong()` is the “legacy” version that is also provided by the old shell.

source/reference/data-types.txt
davemungo: Done

source/reference/data-types.txt
davemungo: Done

source/reference/data-types.txt
davemungo: Done

source/reference/data-types.txt
jeff-allen-mongo: Sounds good. Thanks for the clarifications!

jeff-allen-mongo (approved):

PR #127, opened 2021-07-07
(DOCSP-17438): Use 'built-in' instead of 'native' module

jeff-allen-mongo:
JIRA: [DOCSP-17438](https://jira.mongodb.org/browse/DOCSP-17438)

Staging: [Write Scripts for mongosh](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-17438/write-scripts)

### DIFF
--- a/source/write-scripts.txt
+++ b/source/write-scripts.txt
@@ -157,13 +157,12 @@ You can ``require()`` built-in Node modules (such as
 
       var fs = require('fs');
 
-Require a Non-Built-In Module
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Require an NPM Module
+~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


source/write-scripts.txt
```diff
@@ -157,20 +157,21 @@ You can ``require()`` native Node modules (such as
 
       var fs = require('fs');
 
-Require a Non-Native Module
-~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Require a Non-Built-In Module
```
jeff-allen-mongo, line 160: Would something like "External" be preffered over "Non-Built-In" ? Or perhaps some other name?

addaleax: Yes, I think so, but also you're referencing npm packages here explicitly, so I think just saying "npm modules" is fine 🙂

addaleax (approved):

PR #134, opened 2021-07-16
DOSCP-16084 updates here to match docs repo changes

davemungo:


davemungo:
The bulk of the changes for this jira ticket are in the docs repo. These updates are fixes for problems found here while testing the overall install flow as set out in that repo.

PR #138, opened 2021-08-05
DOCSP-17839 point out that zstd and snappy are not available as compressor options

ianf-mongodb:
…essor options


addaleax (approved):
By the way, the kerberos part is no longer accurate, and we should probably do something about that … @rose-m since you probably know best what’s missing, maybe you could open a DOCSP ticket about it?

PR #139, opened 2021-08-05
DOCSP-16167 port shell customization

davemungo:


source/reference/configure-shell-settings.txt
```diff
@@ -91,21 +91,96 @@ Supported ``property`` parameters
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Use a function like this one to display the database and hostname in
+the :binary:`~bin.mongosh` prompt:
+
+.. code-block:: javascript
+
+   host = db.serverStatus().host;
+   prompt = function() {
+                return db.getName()+"@"+host+"> ";
```
addaleax, line 139:
```suggestion
                return `${db.getName()}@${host}> `;
```

(also just a best practice thing)

It might be worth calling out here that this will update the database name, but not the hostname, if `db` is reassigned (e.g. via `db = connect(...)`).

A version that accounts for that by caching the hostname might look like this:

```js
{
  const hostnameSymbol = Symbol('hostname');
  prompt = () => {
    if (!db[hostnameSymbol]) 
      db[hostnameSymbol] = db.serverStatus().host;
    return `${db.getName()}@${db[hostnameSymbol]}> `;
  };
}
```

I think it’s worth writing it out like this – including the block-scoping and the usage of `const` – so that we encourage users to write idiomatic and reliable code, even if it’s just for prompt customization.

addaleax (approved):

source/reference/configure-shell-settings.txt
```diff
@@ -91,21 +91,96 @@ Supported ``property`` parameters
...
+
+Display Line Numbers
+~~~~~~~~~~~~~~~~~~~~
+
+This code will dynamically update the :binary:`~bin.mongosh` prompt
+to display line numbers:
+
+.. code-block:: javascript
+
+   cmdCount = 1;
```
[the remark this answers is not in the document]
davemungo: Done.

source/reference/configure-shell-settings.txt
davemungo: Thanks Anna, that's very helpful

PR #140, opened 2021-08-05
DOCSP-16438 redacted history is UX pain point

davemungo:
This updates the docs to add info about the modes, a behavior section, and an example for the 'redactHistory' setting

# JIRA:
https://jira.mongodb.org/browse/DOCSP-16438

# STAGING
https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-16438-redacted-history-is-ux-pain-point/reference/configure-shell-settings/

### DIFF
--- a/source/reference/configure-shell-settings.txt
+++ b/source/reference/configure-shell-settings.txt
@@ -1,8 +1,8 @@
 .. _mongosh-shell-settings:
 
-=====================
-Configure ``mongosh``
-=====================
+================================
+Configure :binary:`~bin.mongosh`
+================================
 
 .. default-domain:: mongodb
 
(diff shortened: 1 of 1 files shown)


source/reference/configure-shell-settings.txt
```diff
@@ -176,3 +218,52 @@ API call.
...
+When you press the ``up arrow`` to replay the last command the email
+address is present. 
+
+.. code-block:: javascript
+   :copyable: false
+
+   db.contacts.find( {"email": "customer@clients.com" } )
+
+The :ref:`shell history log <mdb-shell-command-history>` reflects the
+changes. (Note that this log stores the most recent input first.)
```
addaleax, line 260: I think it might be helpful to avoid referring to history “log” files here, because mongosh also maintains log files that are referred to by that name which also contain the command history, even if in a different format

addaleax (approved):

source/reference/configure-shell-settings.txt
davemungo: Updated,

PR #143, opened 2021-08-24
(DOCSP-17607): Add supported mongodb versions to install / connect docs

jeff-allen-mongo:


source/install.txt
```diff
@@ -12,18 +12,6 @@ Install ``mongosh``
    :depth: 1
    :class: singlecol
 
-Prerequisites
```
jeff-allen-mongo, line 15: Note: I ended up removing this section from the Install page and moving it to the Connect page, since I thought it made more sense there. If we want to keep it on the install page as well, we can do so. Just let me know.

addaleax (approved):

PR #144, opened 2021-08-26
DOCSP-17908 Timestamp argument order

davemungo:
Documents the order of parameters for the Timestamp constructor 

JIRA
https://jira.mongodb.org/browse/DOCSP-17908

STAGING
https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-17908-timestamp-argument-order/reference/data-types/#timestamp

### DIFF
--- a/source/reference/data-types.txt
+++ b/source/reference/data-types.txt
@@ -153,16 +153,63 @@ of the binary-based ``Double`` type generally have approximate
 representations of decimal-based values and may not be exactly
 equal to their decimal representations.
 
+Timestamp
+---------
+
+MongoDB uses a
+:manual:`BSON Timestamp </reference/bson-types/#timestamps>` internally
+in the :term:`oplog`. The ``Timestamp`` type works similarly to the
+Java Timestamp type. Use the :ref:`Date <mongo-shell-date-type>` type
+for operations involving dates.
+
+A ``Timestamp`` signature has two optional parameters. 
+
+.. code-block:: javascript
+
+   Timestamp( { "t": <integer>, "i": <integer> } )
+
+.. list-table::
+   :header-rows: 1
+   :widths: 18, 18, 25, 39
+
+   * - Parameter
+     - Type
+     - Default
+     - Definition
+
+   * - ``t``
+     - integer
+     - Current time since :term:`UNIX epoch <unix epoch>`.
+     - Optional. A time in seconds.
+     
+   * - ``i``
+     - integer
+     - 1
+     - Optional. Used for ordering when there are multiple operations
+       within a given second. ``i`` has no effect if used without
+       ``t``.
+
+For usage examples, see :ref:`timestamp-ex-default`,
+:ref:`timestamp-ex-custom`.
+
 .. _check-types-in-shell:
-      
+
 Type Checking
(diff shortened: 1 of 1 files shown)


source/reference/data-types.txt
```diff
@@ -472,9 +516,117 @@ stored as type ``int``.
...
+      }
+   ]
+
+.. _timestamp-ex-extract:
+
+Extract the Date from a System Generated Timestamp
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+MongoDB uses the Timestamp() constructor internally when it creates an
+:ref:`ObjectId`. You can use the :method:`ObjectId.getTimestamp()`
```
addaleax, line 613: The point of my earlier comment was that Timestamp and ObjectId are entirely unrelated types, other than that both use UNIX timestamps (i.e. the first sentence here is inaccurate), and that I think it would be confusing for readers if the docs imply that they are related in some way.

davemungo: I've removed the example, would you have a look? Thanks!

source/reference/data-types.txt
```diff
@@ -153,16 +153,60 @@ of the binary-based ``Double`` type generally have approximate
 representations of decimal-based values and may not be exactly
 equal to their decimal representations.
 
+Timestamp
+---------
+
+MongoDB uses a
+:manual:`BSON Timestamp </reference/bson-types/#timestamps>` internally
+in the :term:`oplog` and as a component of a system generated
+:manual:`ObjectId </reference/bson-types/#objectid>`. The ``Timestamp``
```
addaleax, line 162: This also applies here :)

source/reference/data-types.txt
```diff
@@ -153,16 +153,60 @@ of the binary-based ``Double`` type generally have approximate
...
+
 Type Checking
 -------------
 
-Use the :query:`$type` query operator to determine types. The
-Javascript ``instanceof`` and ``typeof`` operators are not reliable
+Use the :query:`$type` query operator or examine the object constructor
+to determine types. 
+
+The Javascript ``instanceof`` and ``typeof`` operators are not reliable
```
addaleax, line 205: When is `typeof` unreliable?

(Also: `instanceof` being unreliable is not mongosh-specific. You have the same problem in browsers or Node.js.)

davemungo: I think you told me that when we worked on this section a few months ago (https://jira.mongodb.org/browse/DOCSP-16365), but maybe I misread it. The issue with typeof is that it returns the more generic 'object' instead of, say, 'ObjectId'.

addaleax: Right – `typeof` only helps you distinguish between the different kinds of primitives, and objects vs functions. So it’s not very precise, but it is reliable (especially for detecting primitives, which is what it’s typically used for).

addaleax (approved):

source/reference/data-types.txt
davemungo: I've updated the copy for type checking, would you take a look?

PR #145, opened 2021-09-14
(DOCSP-18675): mongosh 1.0.6 release notes

jeff-allen-mongo:


source/changelog.txt
```diff
@@ -12,6 +12,37 @@
...
    :class: singlecol
 
+v1.0.6
+------
+
+*Released September 14, 2021*
+
+New features in this release:
+
+- You can now run :method:`sh.status()` when connected to a ``mongos``.
```
addaleax, line 22: @jeff-allen-mongo This is missing the important word “not”. :)

Fwiw, this feature is TSE-facing. We don’t even document that `sh.status()` takes a second argument (which targets the same use case), so I think we should probably also omit this.

source/changelog.txt
```diff
@@ -12,6 +12,37 @@
...
+
+- ``db.setSecondaryOk()``, ``mongo.setSecondaryOk()``, and
+  ``rs.secondaryOk()`` methods are reintroduced but **deprecated**.
+  These methods alias to ``mongo.setReadPref()``.
+
+- When you enter multiline input into the shell, single-line
+  ``// comments`` are now preserved as ``/* comments */`` in the
+  history entry.
+
+- Adds a new snippet for resuming tokens.
```
addaleax, line 32: “resuming tokens” is not really a thing, and I don’t think I’d note individual snippets here since they can be used independently of the mongosh version

source/changelog.txt
jeff-allen-mongo: Oh no! I read your announcement on Slack too quickly. Thanks for spotting this.

source/changelog.txt
jeff-allen-mongo: Understood. Will remove. Thanks!

PR #146, opened 2021-09-14
(DOCSP-18675): Fix 1.0.6 release notes

jeff-allen-mongo:
Hey @addaleax , could you take a look at these fixes to the 1.0.6 release notes? Thank you!

### DIFF
--- a/source/changelog.txt
+++ b/source/changelog.txt
@@ -19,7 +19,9 @@ v1.0.6
 
 New features in this release:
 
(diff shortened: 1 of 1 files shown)


jeff-allen-mongo:
Oh we wanted to remove the sh.status blurb. One sec...

addaleax (approved):

addaleax:
@jeff-allen-mongo Actually, it’s fine if you leave it in – that was a good example and I didn’t realized that it would work when running against a config server as well :+1:

PR #154, opened 2021-10-13
(DOCSP-15619): Document differences between require() and load()

jeff-allen-mongo:
This update adds a new reference page explaining the differences between require() and load() for shell scripting.

This JIRA ticket was the basis for much of the copy: [DOCSP-15619](https://jira.mongodb.org/browse/DOCSP-15619)

Note that the JIRA ticket has an "Exposing Functionality" section which I've omitted from this CR. I believe this info will be covered in the "Code Scoping" section of the Snippets copy (https://jira.mongodb.org/browse/DOCSP-16500).

Changes are staged here:

https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-15619/require-load-differences/

### DIFF
--- a/source/require-load-differences.txt
+++ b/source/require-load-differences.txt
@@ -1,0 +1,135 @@
+================================================
+Differences Between ``require()`` and ``load()``
+================================================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecol
+
+The ``require()`` and ``load()`` methods include files and modules in
+your scripts for added functionality. However, ``require()`` and
+``load()`` differ in their behaviors and availability.
+
+Types of Scripts in ``mongosh``
+-------------------------------
+
+You can use the following types of scripts with ``mongosh``:
+
+- ``mongosh`` scripts, which can be any of the following:
+
+  - Code entered directly into the REPL.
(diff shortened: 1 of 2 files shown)


source/require-load-differences.txt
```diff
@@ -0,0 +1,131 @@
...
+  file being executed.
+
+  .. example::
+
+     To load a ``mongosh`` script named ``test-suite.js`` from another
+     ``mongosh`` script, add the following line to your script:
+
+     .. code-block:: javascript
+
+       load(__dirname + '/test-suite.js')
```
jeff-allen-mongo, line 92: @addaleax , @davemungo raised a question about this example that we wanted to run by you:

> Why would you prepend the absolute path to the current directory when load() can use relative paths? 

I tried to incorporate the example you provided in the JIRA ticket, so wasn't sure if I maybe misunderstood or if there's something we should adjust about the docs here.

addaleax: The problem is that `load()` inteprets path relative to the current working directory of the `mongosh` process.

That’s fine if you load scripts directly from the shell, because you (proably) know where that working directory is, but if you load a script from another script, you want to be sure that the script you’re loading does not depend on external factors like where `mongosh` was started.

jeff-allen-mongo: That makes sense, thank you for the explanation!

PR #156, opened 2021-10-20
DOCSP-19198 command line file name parameter

davemungo:


source/write-scripts.txt
```diff
@@ -197,15 +202,162 @@ the following calls within the |mdb-shell| are equivalent:
...
+      .. code-block:: javascript
+
+         db = connect( 'mongodb://localhost/films' );
+         printjson( db.movies.find( {} ) ); 
+
+   #. Run the scripts from the command line.
+
+      .. code-block:: javascript
+
+         mongosh --file loadMovies.js -f queryMovies.js
```
addaleax, line 299: Is this supposed to be run with `--nodb`? Otherwise this will fail if no server is running on localhost:27017. I think you generally want to use `--nodb` if you assign `db` at the beginning of your script, or alternatively use something like `db = db.getSiblingDB('films')` instead of the `connect()` call.

(I think in practice the latter might have more benefits, since it means that the script is independent from your specific deployment.)

PR #163, opened 2021-11-05
DOCSP-19177 mongosh output not parsable

davemungo:


source/includes/examples/ex-eval-output.rst
```diff
@@ -0,0 +1,29 @@
+.. code-block:: javascript
+
+   mongosh --quiet  --host rs0/centos1104 --port 27500 \
+           --eval "EJSON.stringify(rs.status().members.map(function(m) \
+                  { return {'id':m._id, 'name':m.name, 'stateStr':m.stateStr} }));" \
+   | jq
```
addaleax, line 6: I would suggest using arrow functions here, since there is some value in brevity with `--eval`:

```suggestion
   mongosh --quiet  --host rs0/centos1104 --port 27500 \
           --eval "EJSON.stringify(rs.status().members.map(m => \
                  ({'id':m._id, 'name':m.name, 'stateStr':m.stateStr})));" \
   | jq 
```

or

```suggestion
   mongosh --quiet  --host rs0/centos1104 --port 27500 \
           --eval "EJSON.stringify(rs.status().members.map(({ _id, name, stateStr }) => \
                  ({ _id, name, stateStr })));" \
   | jq 
```

I also think it might not be necessary to mention `jq` here, `EJSON.stringify` can do nice formatting on its own by specifying an indentation depth:


```suggestion
   mongosh --quiet  --host rs0/centos1104 --port 27500 \
           --eval "EJSON.stringify(rs.status().members.map(({ _id, name, stateStr }) => \
                  ({ _id, name, stateStr })), null, 2);" 
```

(same arguments as [`JSON.stringify()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify))

source/reference/compatibility.txt
```diff
@@ -157,3 +157,14 @@ in ``mongosh`` better align with the types used by the MongoDB Drivers.
    For more information on managing types, refer to the
    :manual:`schema validation overview </core/schema-validation>`.
 
+``--eval`` Behavior
+-------------------
+
+The output of ``mongo --eval`` has a different format than the output
+of ``mongosh --eval``.
```
addaleax, line 164: I feel like this is pretty vague and might leave users unnecessarily worrying about compatibility concerns.

Neither `mongo` nor `mongosh` provide output that is consistently programmatically parseable (e.g. if a BSON-specific object type is used) – it might be worth just being explicit here and saying that mongosh does not quote object keys in its output.

source/includes/examples/ex-eval-output.rst
davemungo: jq is important here because the ticket was to show how to create output that's parsable. We pretty much have to pass it through a parser to demonstrate that wrapping in EJSON works. That said, it's helpful to know about EJSON formatting options so I'll add that in too.

addaleax (approved):

PR #167, opened 2021-11-15
DOCSP-16154 Graceful handling of ctrl+c

davemungo:


source/run-commands.txt
```diff
@@ -117,13 +117,30 @@ using the :option:`--norc <mongosh --norc>` option.
...
 press ``Ctrl + C``.
 
-``Ctrl + C`` terminates the process in the shell, but does not
-terminate the process on the MongoDB server. This behavior
-differs from the :method:`db.killOp()` method, which terminates
-commands on the server.
+When you enter ``Ctrl + C``, :binary:`~bin.mongosh`: 
+
+- interrupts the active command,
+- tries to terminate the running process, and
```
addaleax, line 126: I'm not sure the user (or me, tbh) would be entirely sure on the difference between these two points. Maybe instead of "running process", we could say "ongoing server-side operation" here? ("terminate a process" sounds like killing the shell process itself, which is not what happens)

addaleax (approved):

source/run-commands.txt
davemungo: Done

PR #171, opened 2021-11-24
DOCSP-19860 add config.set info for VSCode

davemungo:


source/reference/editor-mode.txt
```diff
@@ -246,10 +246,19 @@ Using Visual Studio as an External Editor
...
 .. code-block:: javascript
 
    export EDITOR="/usr/local/bin/code --wait"
 
+Set the Visual Studio editor with
+:ref:`config.set() <mongosh-shell-settings>`:
+
+.. code-block:: javascript
+
+   config.set("editor", "code --wait")
```
ianf-mongodb, line 260: Hey Dave, this looks good and in my experience worked much better than the absolute path.

Only thing I would add is a note that specifies that "code" must be specified as an alias or configured with your terminal for "code" to work. I remember "code ." did not work straight out of the box for me, something like :  

.. note::
   
   If using the code alias visual studio code needs to be configured as an alias for your terminal or command line   environment.

ianf-mongodb (commented):
Hi Dave, one suggestion for this edit.

source/reference/editor-mode.txt
davemungo: Done.

source/reference/editor-mode.txt
```diff
@@ -246,10 +246,20 @@ Using Visual Studio as an External Editor
...
 editor. Use ``--wait`` with Visual Studio.
 
+Set an environment variable:
+
 .. code-block:: javascript
 
    export EDITOR="/usr/local/bin/code --wait"
 
+You can also set the editor with :ref:`config.set()
+<mongosh-shell-settings>`. If Visual Studio is in your path, open
```
addaleax, line 256: nit: Does it make sense to capitalize `path` to `PATH`? It’s how the environment variable is commonly spelled and `your path` might sound a bit unclear for users who don’t know what it refers to.

ianf-mongodb: Suggestion to update this:

"If Visual Studio is in the system ``PATH`` "

Similar context is used in : 
https://docs.mongodb.com/manual/core/security-automatic-client-side-encryption/#enabling-automatic-client-side-field-level-encryption

source/reference/editor-mode.txt
ianf-mongodb: LGTM!

ianf-mongodb (approved):

PR #188, opened 2022-02-24
DOCSP-16135 match server support matrix

davemungo:
The jira ticket says it is ok to remove the note about mongocryptd. However, there is existing copy about using older platforms and mongosh is supported back to MongoDB 4.0. With that in mind I modified the copy to highlight dropping support for ubuntu 16.04 instead of removing the information entirely. 

### STAGING
https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-16135-matchserver-support-matrix/install/#considerations-for-ubuntu

### JIRA
https://jira.mongodb.org/browse/DOCSP-16135
[MONGOSH] Match server support matrix for OS/arch

### DIFF
--- a/source/install.txt
+++ b/source/install.txt
@@ -108,6 +108,10 @@ Select the appropriate tab for your operating system:
          .. tab:: .deb
             :tabid: deb
 
+            :binary:`mongosh` is available as a :abbr:`PPA (Personal
+            Package Archive)` for Ubuntu 20.04 (Focal) and Ubuntu 18.04
+            (Bionic).
+
             .. include:: /includes/steps/install-shell-linux-deb.rst
 
          .. tab:: .rpm
--- a/source/install.txt
+++ b/source/install.txt
@@ -117,8 +117,8 @@ Select the appropriate tab for your operating system:
          .. tab:: .rpm
             :tabid: rpm
             
-            mongosh is available as ``yum`` package for :abbr:`RHEL
-            Red Hat Enterprise Linux)` and Amazon Linux 2.
+            :binary:`mongosh` is available as ``yum`` package for
+            :abbr:`RHEL Red Hat Enterprise Linux` and Amazon Linux 2.
 
             .. include:: /includes/steps/install-shell-linux-rpm.rst
 
--- a/source/install.txt
+++ b/source/install.txt
@@ -105,7 +105,7 @@ Select the appropriate tab for your operating system:
       - The ``mongocryptd`` binary that is included in the |mdb-shell|
         ``.deb`` package does not work on unsupported Ubuntu platforms
         (16.04 or older). If you must install :binary:`mongosh` on an
-        unsupported Ubuntu platform :manual:`install
+        unsupported Ubuntu platform, :manual:`install
         </reference/security-client-side-encryption-appendix/#installation>` 
         the ``mongocryptd`` binary separately to use its 
         functionality in the |mdb-shell|.
--- a/source/install.txt
+++ b/source/install.txt
@@ -57,21 +57,21 @@ Select the appropriate tab for your operating system:
          `Homebrew Website <https://docs.brew.sh/Installation>`__.
 
       The Homebrew package manager is the recommended installation
-      method for ``mongosh`` on macOS. To learn how to manually install
-      ``mongosh`` from an archive instead, see
+      method for :binary:`mongosh` on macOS. To learn how to manually
+      install :binary:`mongosh` from an archive instead, see
       :ref:`macos-install-archive`.
 
       Considerations
       ``````````````
 
-      ``mongosh`` installed with Homebrew does not support
+      :binary:`mongosh` installed with Homebrew does not support
       :manual:`automatic client-side field level encryption
       </core/security-automatic-client-side-encryption/>`.
 
       Procedure
       `````````
 
-      To install ``mongosh`` with Homebrew:
+      To install :binary:`mongosh` with Homebrew:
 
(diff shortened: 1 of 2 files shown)


source/install.txt
```diff
@@ -91,21 +91,22 @@ Select the appropriate tab for your operating system:
       Considerations for Ubuntu
       ~~~~~~~~~~~~~~~~~~~~~~~~~
 
-      - ``mongosh`` is available as a
+      - :binary:`mongosh` is available as a
         :abbr:`PPA (Personal Package Archive)` for Ubuntu 20.04 (Focal)
-        and Ubuntu 18.04 (Bionic).
+        and Ubuntu 18.04 (Bionic). If you are installing on Ubuntu
+        20.04 (Focal) or Ubuntu 18.04  (Bionic), use the procedure in
```
jeff-allen-mongo, line 97: Looks like there's extra space between 18.04 and Bionic.

source/install.txt
```diff
@@ -91,21 +91,22 @@ Select the appropriate tab for your operating system:
...
+        the ``.deb`` tab.
```
jeff-allen-mongo, line 98: It seems redundant to tell users which tab to click both in the "Considerations for Ubuntu" section and in the "Procedure" section. Maybe we can clean this up a bit?

source/install.txt
```diff
@@ -91,21 +91,22 @@ Select the appropriate tab for your operating system:
...
+        tab.
       
-      - The ``mongocryptd`` binary in the |mdb-shell| ``.deb`` 
-        package does not work on Ubuntu 16.04 or older. On 
-        Ubuntu 16.04 and older systems, you must 
-        :manual:`install </reference/security-client-side-encryption-appendix/#installation>` 
+      - The ``mongocryptd`` binary that is included in the |mdb-shell|
+        ``.deb`` package does not work on unsupported Ubuntu platforms
+        (16.04 or older). If you must install :binary:`mongosh` on an
+        unsupported Ubuntu platform, :manual:`install
```
jeff-allen-mongo, line 108: [suggestion]

When only the "install" text is hyperlinked, it's a little unclear where this link will go. Maybe we should remove the hyperlink from install and add a sentence to this end of this bullet like 

```
To learn how to install ``mongocrptd``, see :manual:`installation
</reference/security-client-side-encryption-appendix/#installation>`
```

source/install.txt
```diff
@@ -91,21 +91,22 @@ Select the appropriate tab for your operating system:
...
-        - If you are using Ubuntu 20.04 (Focal) or Ubuntu 18.04
-          (Bionic), use the procedure in the ``.deb`` tab.
-        
-        - If you are using Ubuntu 16.04 (Xenial), you must
-          obtain the ``mongosh`` tarball using the procedure in the
-          ``.tgz`` tab.
+      - MongoDB 5.0 removes support for Ubuntu 16.04 (Xenial). If you
+        must use the unsupported Ubuntu 16.04 (Xenial), install the
+        :binary:`mongosh` tarball using the procedure in the ``.tgz``
+        tab.
```
jeff-allen-mongo, line 103: [question] Should we be telling users how to install mongosh on unsupported platforms? It's a little confusing to me why we aren't just telling them they can't install mongosh on Xenial.

jeff-allen-mongo (commented):
Thanks @davemungo , I've left a few questions and suggestions.

davemungo (commented):
@jeff-allen-mongo   I've removed the instructions for unsupported platforms and cleaning up the other pre-existing copy too. Would you have a look?

source/install.txt
```diff
@@ -80,35 +80,14 @@ Select the appropriate tab for your operating system:
...
-          ``.tgz`` tab.
-      
-      - The ``mongocryptd`` binary in the |mdb-shell| ``.deb`` 
-        package does not work on Ubuntu 16.04 or older. On 
-        Ubuntu 16.04 and older systems, you must 
-        :manual:`install </reference/security-client-side-encryption-appendix/#installation>` 
-        the ``mongocryptd`` binary separately to use its 
-        functionality in the |mdb-shell|.
-
       Procedure
```
jeff-allen-mongo, line 91: It may still be valuable to include a section immediately preceding "Procedure" to list out all of the supported Linux architectures. This way, users can tell at a glance whether their architecture is supported.

Under the `.deb` tab, we mention which platforms users can install on. We say:

> :binary:`mongosh` is available as a :abbr:`PPA (Personal
   Package Archive)` for Ubuntu 20.04 (Focal) and Ubuntu 18.04 (Bionic).

It might make sense to pull that info up higher on the page and create a dedicated section that says "here are the platforms on which you can run monosh".

Something like:

```
Supported Linux Platforms
~~~~~~~~~~~~~~~~~~~~

You can run mongosh on...

- Ubuntu...
- RHEL...
- Amazon...
```

jeff-allen-mongo (approved):
LGTM, but I left a consideration.

source/install.txt
davemungo: The PPA info is needed here because we need to let users know that mongosh isn't in the Ubuntu repo. 

I'll open another ticket for a more general, supported platform list.

davemungo: https://jira.mongodb.org/browse/DOCSP-21075

davemungo:
@addaleax  This ticket removes discussion of mongocryptd and the unsupported Ubuntu 16.04. Would you have a look when you get a moment?

addaleax (approved):

PR #194, opened 2022-03-08
DOCSP-21243 retryable writes enabled by default

davemungo:
These are updates in the mongosh repo to accompany updates in the server manual repo (DOCSP-21096)

### STAGING
https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-21243-retryable-writes-enabled-by-default/reference/compatibility/#write-preference-behavior

https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-21243-retryable-writes-enabled-by-default/reference/options/#std-option-mongosh.--retryWrites

### JIRA
https://jira.mongodb.org/browse/DOCSP-21243
[MONGOSH] Retryable writes enabled by default

### DIFF
--- a/source/reference/compatibility.txt
+++ b/source/reference/compatibility.txt
@@ -112,6 +112,14 @@ specified for the operation:
 In the legacy ``mongo`` shell, these operations use the specified read
 preference.
 
+Write Preference Behavior
+-------------------------
+
+:ref:`Retryable writes <retryable-writes` are the default in
+:binary:`mongosh`. Retryable writes were disabled by default in the
+legacy :binary:`mongo` shell. To disable retryable writes, use 
+:option:`--retryWrites=false <mongosh --retryWrites>`.
+
 Numeric Values
 --------------
 
--- a/source/reference/compatibility.txt
+++ b/source/reference/compatibility.txt
@@ -115,7 +115,7 @@ preference.
 Write Preference Behavior
 -------------------------
 
-:ref:`Retryable writes <retryable-writes>` are the default in
+:ref:`Retryable writes <retryable-writes>` are enabled by default in
 :binary:`mongosh`. Retryable writes were disabled by default in the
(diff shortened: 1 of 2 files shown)


jason-price-mongodb (approved):
@davemungo LGTM. Cheers.

source/reference/compatibility.txt
```diff
@@ -112,6 +112,14 @@ specified for the operation:
 In the legacy ``mongo`` shell, these operations use the specified read
 preference.
 
+Write Preference Behavior
+-------------------------
+
+:ref:`Retryable writes <retryable-writes>` are the default in
```
Anemy, line 118: How about this to be a bit more transparent that the default is `on/true`
```suggestion
:ref:`Retryable writes <retryable-writes>` are enabled by default in
```

source/reference/options.txt
```diff
@@ -452,8 +452,10 @@ Session Options
 
 .. option:: --retryWrites
 
-   Enables :manual:`retryable writes </core/retryable-writes/>` as the
-   default for sessions in the |mdb-shell|.
+   Enables :ref:`retryable-writes`. Retryable writes are the default in
```
Anemy, line 455: Same suggestion from above - how about this to be a bit more transparent that the default is `on/true`?
```suggestion
   Enables :ref:`retryable-writes`. Retryable writes are enabled by default in
```

Anemy (approved):
lgtm! Two suggestions, fine as is as well.

source/reference/options.txt
davemungo: Done

source/reference/compatibility.txt
davemungo: Done

davemungo (commented):
Thanks @Anemy  updated as suggested, merging next.

PR #195, opened 2022-03-10
(DOCSP-21310,DOCSP-21055): 1.2.2, 1.2.3 release notes

jeff-allen-mongo:


source/changelog.txt
```diff
@@ -12,6 +12,36 @@
...
+*Released March 10, 2022*
+
+New features:
+
+- :issue:`MONGOSH-1121` - Support the ``commitQuorum`` parameter for the
+  :method:`~db.collection.createIndexes()` method.
+
+Bug fixes:
+
+- Use Rocky Linux 8 instead of CentOS 7 for packaging.
```
addaleax, line 27: This doesn’t affect users – if there’s anything to call out here, it’s that .rpm packages now have SHA256 digests, but 99.9 % of users won’t care about that either :)

jeff-allen-mongo: Got it, will remove 👍

PR #199, opened 2022-03-30
DOCSP-15571 productize new async writer

davemungo:
### STAGING
https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-15571-productize-new-async-rewriter/reference/compatibility/#limitations-on-database-calls
https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-15571-productize-new-async-rewriter/write-scripts/

### JIRA
https://jira.mongodb.org/browse/DOCSP-15571
[MONGOSH] Productize new async rewriter

### DIFF
--- a/source/includes/fact-call-limitations.rst
+++ b/source/includes/fact-call-limitations.rst
@@ -48,6 +48,22 @@ Use an ``async`` function instead:
       }
    }
 
+.. note::
+
+   You can also use a method that performs a database operation inside
+   a class as an alternative to working with asynchronous JavaScript.
+
+   .. code-block::
+
+      class FindResults {
+         constructor() { }
+         
+         init() { this.value = db.students.find(); }
+       }
+
+   To use this class, first construct a class instance then call the
+   ``.init()`` method.
+
 Generator Functions
 ~~~~~~~~~~~~~~~~~~~
 
--- a/source/includes/fact-call-limitations.rst
+++ b/source/includes/fact-call-limitations.rst
@@ -96,7 +96,7 @@ Use an ``async generator function`` instead:
 Array Sort
 ~~~~~~~~~~
 
-The following array sort do not work: 
+The following array sort does not work: 
 
 .. code-block:: javascript
    :copyable: false
--- a/source/includes/fact-call-limitations.rst
+++ b/source/includes/fact-call-limitations.rst
@@ -1,0 +1,104 @@
+.. Top level title in calling page, won't render in right TOC
+
+The results of database calls cannot be passed inside the following
+contexts:
+
+- Class constructor functions
+- Non-async generator functions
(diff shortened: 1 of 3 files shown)


source/includes/fact-call-limitations.rst
```diff
@@ -0,0 +1,104 @@
...
+To access to the results of database calls, use `async functions
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function>`__,
+`async generator functions
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of>`__,
+or ``.map()``.
+
+Constructors
+~~~~~~~~~~~~
+
+The following constructors will not work:
```
jmd-mongo, line 19: [suggestion]

Consider using the present tense.

jmd-mongo (commented):
Thanks for the changes, @davemungo! I've left some comments and questions for you.

Thanks!
Joe

davemungo:
@addaleax Would you have a look at these updates?

Two technical questions - 
Is it ok to say, "database queries" instead of "database calls"?

The examples use a find() query. Does it make sense to show an example with some other sort of query? 

Thanks!

source/includes/fact-call-limitations.rst
```diff
@@ -0,0 +1,104 @@
...
+
+.. code-block:: javascript
+
+   class FindResults {
+      constructor() {
+         this.value = ( async() => {
+            return db.students.find();
+         } )();
+      }
+   }
```
addaleax, line 49: It *might* also be worth adding a second alternative here, e.g.:

```js
class FindResults {
  constructor() {
  }

  init() {
    this.value = db.students.find();
  }
}
```

This would mean that consumers of this class would have to call `.init()` after already having constructed the class instance, but the big advantage would be that there is no knowledge of asynchronous code in JS required in order for this to work

addaleax (approved):
> Is it ok to say, "database queries" instead of "database calls"?

I think you can do that – but keep in mind that this applies to anything that interacts with the database in some way (for example, commands rather than queries).

> The examples use a find() query. Does it make sense to show an example with some other sort of query?

I don’t think you’d need to provide explicit examples here – the limitations and workarounds would apply identically to any other database call (or database query :slightly_smiling_face:)

davemungo (commented):
@addaleax Thanks for your feedback. I added a short note about the aysnch alternative.

jason-price-mongodb (changes requested):
@davemungo please address mistake and resubmit PR. Cheers.

davemungo (commented):
Thanks @jason-price-mongodb  

I've updated "do not" to "does not"

PR #200, opened 2022-04-04
DOCSP-20366

sarah-olson-mongodb:
Jira: https://jira.mongodb.org/browse/DOCSP-20366
Staging: https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-20366/
What has changed: Updated the snooty.toml file, and edited a few links throughout. 

Two philosophical questions for Jeff:
1. I edited a link to the Versioned/Stable API in the change log. I retained "Versioned API" rather than change it "Stable API" on the grounds that when that specific change was introduced, the name was still Versioned API and changelogs are more like historical record. Let me know if you disagree with this approach. 
https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-20366/changelog/#v0.15.1

2. I updated the links shown in help and API messages even though the actual messages may not be changed for a bit. It probably doesn't matter too much, but I figured it was better to update the links than not. Let me know if you have any concerns here.


source/changelog.txt
```diff
@@ -403,8 +403,7 @@ New features in this release:
 - Autocomplete is now aware of the ``--apiStrict`` flag. When
   ``--apiStrict`` is ``true``, autocomplete only completes methods
   that work with your defined API version. For more information, see
-  `Versioned API
-  <https://docs.mongodb.com/v5.0/reference/versioned-api/>`__.
+  :manual:`Versioned API </reference/stable-api/>`.
```
jeff-allen-mongo, line 406: I think the link text should be "Stable API". Justification in my mind is that users will click this link and arrive on a page called "Stable API". If the link text doesn't match the target page text, users might get confused or think they've arrived in the wrong place.

source/connect.txt
```diff
@@ -220,8 +220,7 @@ Connect to a Specific Database
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 To connect to a specific database, specify a database in your
-`connection string URI path
-<https://docs.mongodb.com/manual/reference/connection-string/>`__. If
+:manual:`connection string URI path</reference/connection-string/>` If
```
jeff-allen-mongo, line 223:
```suggestion
:manual:`connection string URI path </reference/connection-string/>` If
```

source/changelog.txt
jeff-allen-mongo, line 406:
```suggestion
  :manual:`Stable API </reference/stable-api/>`.
```

jeff-allen-mongo (approved):
LGTM % a couple suggestions. Thanks for these updates!

source/changelog.txt
sarah-olson-mongodb: Thanks Jeff! Solid justification; updated! :-)

schmalliso (approved):

PR #228, opened 2022-08-01
(DOCSP-24131): v1.5.4 release notes

jeff-allen-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-24131

Staging: 

https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-24131/changelog/#v1.5.4

### DIFF
--- a/source/changelog.txt
+++ b/source/changelog.txt
@@ -21,7 +21,7 @@ Fixes a potential data corruption bug in
 ``KeyVault.rewrapManyDataKey()`` when rotating encrypted data encryption
 keys backed by Azure or GCP key services.
 
-In previous versions of ``mongosh``, this bug occurrs when an
+In previous versions of ``mongosh``, this bug occurs when an
 Azure-backed or GCP-backed data encryption key being rewrapped required
 fetching an access token for decryption of the data encryption key.
 
--- a/source/changelog.txt
+++ b/source/changelog.txt
@@ -18,11 +18,11 @@ v1.5.4
 *Released July 31, 2022*
 
 Fixes a potential data corruption bug in
-``KeyVault.rewrapManyDataKey()`` when rotating encrypted data encryption
-keys backed by Azure or GCP key services.
+:method:`KeyVault.rewrapManyDataKey()` when rotating encrypted data
+encryption keys backed by Azure or GCP key services.
 
 In previous versions of ``mongosh``, this bug occurs when an
-Azure-backed or GCP-backed data encryption key being rewrapped required
+Azure-backed or GCP-backed data encryption key being rewrapped requires
 fetching an access token for decryption of the data encryption key.
 
 As a result of this bug, all data encryption keys being rewrapped are
@@ -30,7 +30,7 @@ replaced by new randomly generated material, destroying the original key
 material.
 
 To mitigate potential data corruption, upgrade ``mongosh`` to v1.5.4 or
-higher before using ``KeyVault.rewrapManyDataKey()`` to rotate
+higher before using :method:`KeyVault.rewrapManyDataKey()` to rotate
 Azure-backed or GCP-backed data encryption keys. You should always
 create a backup of the key vault collection before key rotation.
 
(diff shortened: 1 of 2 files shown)


source/changelog.txt
```diff
@@ -12,6 +12,28 @@ Release Notes
...
+v1.5.4
+------
+
+*Released July 31, 2022*
+
+Fixes a potential data corruption bug in
+``KeyVault.rewrapManyDataKey()`` when rotating encrypted data encryption
+keys backed by Azure or GCP key services.
+
+In previous versions of ``mongosh``, this bug occurrs when an
```
addaleax, line 24:
```suggestion
In previous versions of ``mongosh``, this bug occurs when an
```

addaleax (approved):

source/changelog.txt
```diff
@@ -12,6 +12,28 @@ Release Notes
...
+Azure-backed or GCP-backed data encryption key being rewrapped required
```
davemungo, line 25: I know this is the suggested text, but we should at least fix the tenses:

Either:
in previous versions of ``mongosh``, this bug occured when an Azure-backed or GCP-backed data encryption key that was being rewrapped required fetching an access token for decryption of the data encryption key.

Or:
n previous versions of ``mongosh``, this bug occurs when an Azure-backed or GCP-backed data encryption key that is being rewrapped requires fetching an access token for decryption of the data encryption key.

source/changelog.txt
```diff
@@ -12,6 +12,28 @@ Release Notes
...
+fetching an access token for decryption of the data encryption key.
+
+As a result of this bug, all data encryption keys being rewrapped are
```
davemungo, line 28: What does "material" mean here? Are the keys replaced by random cypher text or is the underlying data replaced by random cypher text? (You could potentially recover from the first problem. It would be much harder to recover from not the second issue.)

source/changelog.txt
```diff
@@ -12,6 +12,28 @@ Release Notes
    :depth: 1
    :class: singlecol
 
+v1.5.4
+------
+
+*Released July 31, 2022*
+
+Fixes a potential data corruption bug in
+``KeyVault.rewrapManyDataKey()`` when rotating encrypted data encryption
```
davemungo, line 21: Should this be a link?

davemungo (commented):
@jeff-allen-mongo  You're working with a suggested text so I understand that options are limited, but I've left a couple of suggestions to consider.

source/changelog.txt
jeff-allen-mongo: Fixed

source/changelog.txt
jeff-allen-mongo: @addaleax are you able to answer this question, and also if we should adjust the text to be clearer about what's meant by "material"?

addaleax: > What does "material" mean here?

It refers to the plaintext of the data encryption key, something that the user or the database never interacts with directly.

libmongocrypt creates data keys by generating random data (= key material), then encrypting it using the KMS and storing it in the key vault. When user data is encrypted/decrypted, it fetches the data key from the key vault, decrypts its contents using the KMS to get the key material, then encrypts/decrypts the data using that key material. With this bug, the key material is replaced with random data, leaving no way to decrypt user data.

> also if we should adjust the text to be clearer about what's meant by "material"?

These release notes were user for driver and libmongocrypt releases, and the wording here has been signed off on by product/drivers leads. I’d not worry about the exact wording here in the docs too much.

jeff-allen-mongo: Thanks for the clarification Anna. I'll keep the wording here as-is.

source/changelog.txt
jeff-allen-mongo: Added.

jeff-allen-mongo:
@davemungo thanks for the review. Please have another look.

source/changelog.txt
davemungo: Understood, I just wish they would have run the text past one of us first.

davemungo (approved):
@jeff-allen-mongo  this LGTM

PR #229, opened 2022-08-02
DOCSP-23367 multiple eval args

davemungo:
[STAGING](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-23367-multiple-eval-args/reference/options/#std-option-mongosh.--eval)

[JIRA](https://jira.mongodb.org/browse/DOCSP-23367)

### DIFF
--- a/source/includes/examples/ex-eval-output.rst
+++ b/source/includes/examples/ex-eval-output.rst
@@ -1,12 +1,12 @@
 To get output suitable for automated parsing, use
 ``EJSON.stringify()``.
 
- .. code-block:: javascript
+.. code-block:: javascript
 
-   mongosh --quiet  --host rs0/centos1104 --port 27500 \
-           --eval "EJSON.stringify(rs.status().members.map( \
-             m => ({'id':m._id, 'name':m.name, 'stateStr':m.stateStr})));" \
-   | jq 
+  mongosh --quiet  --host rs0/centos1104 --port 27500 \
+          --eval "EJSON.stringify(rs.status().members.map( \
+            m => ({'id':m._id, 'name':m.name, 'stateStr':m.stateStr})));" \
+  | jq 
 
 After parsing with ``jq``, the output resembles this:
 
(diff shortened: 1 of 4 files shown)


source/reference/options.txt
```diff
@@ -25,19 +25,22 @@ General Options
...
+   argument or multiple ``--eval`` arguments together.
+   
+   After ``mongosh`` evaluates the ``--eval`` argument, it prints the
+   results to your command line. If you use multiple ``--eval``
+   statements, ``mongosh`` only prints the results of the last
+   ``--eval``.
 
-   To get output suitable for automated parsing, use
-   ``EJSON.stringify()``.
+   **Example: Format Output**
```
jmd-mongo, line 36: [optional suggestion] 

Consider using the "Example" admonition for these two cases.

jmd-mongo (approved):
Thanks for the updates, @davemungo. This one LGTM with an optional suggestion.

Thanks!
Joe

source/reference/options.txt
davemungo: Thanks @jmd-mongo

This might be a good spot for it, but I think we agreed not to use the "example" since it can be distracting on the page, especially in this case where the existing example is quite long.

davemungo:
@addaleax would you have a look at this one when you get a moment?

source/includes/examples/ex-eval-output.rst
```diff
@@ -1,9 +1,12 @@
+To get output suitable for automated parsing, use
+``EJSON.stringify()``.
```
addaleax, line 2: Just a heads up, no need to change anything – while this is perfectly true and correct advice, we are planning to provide a less cumbersome way of getting parse-able results as part of [MONGOSH-1249](https://jira.mongodb.org/browse/MONGOSH-1249)).

addaleax (approved):

source/includes/examples/ex-eval-output.rst
davemungo: Cool, thanks for the info

PR #233, opened 2022-08-25
DOCSP-24855 adds Stable API options

jmd-mongo:
# STAGING

[Stable API Options](https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-24855/reference/options/#stable-api-options)

# JIRA

https://jira.mongodb.org/browse/DOCSP-24855

### DIFF
--- a/source/reference/options.txt
+++ b/source/reference/options.txt
@@ -10,36 +10,18 @@ Options
    :depth: 2
    :class: singlecol
 
+.. program:: mongosh
+
 Use the following options to control various aspects of your
 |mdb-shell| connection and behavior.
 
 General Options
 ---------------
 
-.. program:: mongosh
-
 .. option:: --build-info
 
    Returns a JSON-formatted document with information about the
    :binary:`~bin.mongosh` build.
-
-.. option:: --eval <javascript>
-
-   Evaluates a JavaScript expression. You can use a single ``--eval``
-   argument or multiple ``--eval`` arguments together.
-   
-   After ``mongosh`` evaluates the ``--eval`` argument, it prints the
-   results to your command line. If you use multiple ``--eval``
-   statements, ``mongosh`` only prints the results of the last
-   ``--eval``.
-
-   **Example: Format Output**
-
-   .. include:: /includes/examples/ex-eval-output.rst
-
-   **Example: Multiple ``--eval`` Arguments**
-
-   .. include:: /includes/examples/ex-eval-multi.rst
 
 .. option:: --help, -h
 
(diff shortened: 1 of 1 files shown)


source/reference/options.txt
```diff
@@ -76,6 +76,34 @@ General Options
...
+   currently the only supported value.
+
+.. option:: --apiStrict
+
+   Specifies that the server will respond with :ref:`APIStrictError 
+   <api-strict-resp>` if your application uses a command or behavior 
+   outside of the :ref:`Stable API <stable-api>`.
+   
+   If you specify :option:`--apiStrict`, you must also specify
+   :option:`--apiVersion`.
```
sarah-olson-mongodb, line 96: (question) Is this for apiStrict? From the description, it seems at least also intended for apiDeprecationErrors, but perhaps only intended?

source/reference/options.txt
```diff
@@ -76,6 +76,34 @@ General Options
...
+
+.. option:: --apiDeprecationErrors
+
+   Specifies that the server will respond with :ref:`APIDeprecationError
+   <api-deprecation-resp>` if your application uses a command or behavior 
+   that is deprecated in the specified :ref:`apiVersion
```
sarah-olson-mongodb, line 102: (linking)
This is the second instance of the apiVersion link. Fine to omit.

sarah-olson-mongodb (approved):
LGTM.
Thanks for this work Joe!
I left two items for your consideration.

source/reference/options.txt
jmd-mongo: Good eye! I've noted this in the description for --apiDeprecationErrors as well.

source/reference/options.txt
jmd-mongo: Omitted.

source/reference/options.txt
```diff
@@ -76,6 +76,36 @@ General Options
...
+.. option:: --apiVersion <version number>
+
+   Specifies the :ref:`apiVersion <api-version-desc>`. ``"1"`` is 
+   currently the only supported value.
+
+.. option:: --apiStrict
+
+   Specifies that the server will respond with :ref:`APIStrictError 
+   <api-strict-resp>` if your application uses a command or behavior 
+   outside of the :ref:`Stable API <stable-api>`.
```
Anemy, line 93: Should we mention that when this is not specified then mongosh defaults to a non-strict client? I think it's probably fine as is as well.

Anemy (approved):
lgtm - left one thought/suggestion.

PR #242, opened 2022-12-02
DOCSP-26805 mongosh 1.6.1 release notes

jmd-mongo:
## STAGING

https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-26805/changelog/#v1.6.1

## JIRA

https://jira.mongodb.org/browse/DOCSP-26805

## BUILD LOG

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=638a31fbed8015c14b367660

### DIFF
--- a/source/changelog.txt
+++ b/source/changelog.txt
@@ -20,9 +20,6 @@ v1.6.1
 - :issue:`MONGOSH-1320`: Fixes a startup bug related to Docker and 
   similar environments.
 
-- :issue:`MONGOSH-955`: ``printjson()`` now prints full, untruncated 
-  output.
-
 - :issue:`MONGOSH-1050`: Adds support for the 
   ``convertShardKeyToHashed()`` helper method.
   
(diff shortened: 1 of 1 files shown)


jason-price-mongodb (approved):
@jmd-mongo LGTM. Cheers.

source/changelog.txt
```diff
@@ -12,6 +12,23 @@ Release Notes
...
+v1.6.1
+------
+
+*Released December 1, 2022*
+
+- :issue:`MONGOSH-1320`: Fixes a startup bug related to Docker and 
+  similar environments.
+
+- :issue:`MONGOSH-955`: ``printjson()`` now prints full, untruncated 
+  output.
```
addaleax, line 24: We don't document this method, and I wouldn't necessarily include this item here, as I'm not really convinced that it's a good idea to actually advocate its usage (it's called `printjson` and it doesn't print JSON, for starters). It's only present for legacy shell compatibility, and was included in the Slack announcement since some internal users feel strongly about its behavior.

jmd-mongo: Thanks for the explanation. I've removed this item.

PR #243, opened 2022-12-02
(DOCSP-13713): Improve layout of connect page

jeff-allen-mongo:
## DESCRIPTION

The goal here is to remove the clunky ``.. example::`` admonitions on this page and replace them with in-line examples. I also cleaned up some of the wording where appropriate.

## STAGING

https://docs-mongodbcom-staging.corp.mongodb.com/mongodb-shell/docsworker-xlarge/DOCSP-13713/connect/

## JIRA

https://jira.mongodb.org/browse/DOCSP-13713

## BUILD LOG

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=638a38f2ed8015c14b37d44f

### DIFF
--- a/source/connect.txt
+++ b/source/connect.txt
@@ -148,7 +148,7 @@ To connect to a replica set, perform one of the following actions:
 
      mongosh "mongodb://mongodb0.example.com.local:27017,mongodb1.example.com.local:27017,mongodb2.example.com.local:27017/?replicaSet=replA"
 
-.. note::
+.. note:: directConnection Parameter
 
    ``mongosh`` adds the ``directConnection=true`` query parameter to the
    connection string automatically unless at least one of the following is
--- a/source/connect.txt
+++ b/source/connect.txt
@@ -175,7 +175,8 @@ run the following command:
    - The connection string uses the ``mongodb+srv://`` connection string
      format.
    - The connection string contains a seed list with multiple hosts.
-   - The connection string already contains a `directConnection` parameter.
+   - The connection string already contains a ``directConnection``
+     parameter.
 
    When ``directConnection=true``, all operations are run on the host
    specified in the connection URI.
--- a/source/connect.txt
+++ b/source/connect.txt
@@ -121,7 +121,7 @@ Connect to a Replica Set
 
 To connect to a replica set, perform one of the following actions:
 
-- If you are using the :manual:`DNS Seedlist Connection Format
+- If you use the :manual:`DNS Seedlist Connection Format
   </reference/connection-string/#dns-seedlist-connection-format>`,include
   the ``+srv`` modifier in your connection string.
 
--- a/source/connect.txt
+++ b/source/connect.txt
@@ -110,7 +110,9 @@ database, run the following command:
    mongosh "mongodb://mongodb0.example.com:28015" --username alice --authenticationDatabase admin
    
 To provide a password as part of the connection command instead of using
-the masked prompt, use the :option:`--password <--password>` option.
+the prompt, use the :option:`--password <--password>` option. Use this
+option for programmatic usage of ``mongosh``, like a :driver:`driver
+</>`.
 
 .. seealso::
 
--- a/source/connect.txt
+++ b/source/connect.txt
@@ -64,9 +64,8 @@ To specify a remote host and port, you can use either:
 
 - A :manual:`connection string </reference/connection-string>`.
 
-- The ``--host`` and ``--port`` command-line options. If you do not
-  include the ``--port`` option, ``mongosh`` uses the default port
-  27017.
+- The ``--host`` and ``--port`` command-line options. If you omit the
+  ``--port`` option, ``mongosh`` uses the default port 27017.
 
 For example, the following commands connect to a MongoDB deployment
 running on host ``mongodb0.example.com`` and port 28015:
@@ -177,8 +176,7 @@ replica set named ``replA``:
 Connect Using TLS
 ~~~~~~~~~~~~~~~~~
 
-To connect to a deployment using TLS, perform one of the following
-actions:
+To connect to a deployment using TLS, you can either:
 
 - Use the :manual:`DNS Seedlist Connection Format
   </reference/connection-string/#dns-seedlist-connection-format>`. The
@@ -192,7 +190,7 @@ actions:
 
      mongosh "mongodb+srv://server.example.com/"
 
-- Specify the :option:`--tls <--tls>` option to ``true`` in the connection
+- Set the :option:`--tls <--tls>` option to ``true`` in the connection
   string.
 
   For example, run the following command to enable ``tls`` with a
@@ -215,7 +213,7 @@ Connect to a Specific Database
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 To connect to a specific database, specify a database in your
-:manual:`connection string URI path </reference/connection-string/>` If
+:manual:`connection string URI path </reference/connection-string/>`. If
 you do not specify a database in your URI path, you connect to the
 ``test`` database.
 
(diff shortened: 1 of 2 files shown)


source/connect.txt
```diff
@@ -38,133 +37,118 @@ This is equivalent to the following command:
...
 
-     To connect to a MongoDB
-     instance running on a remote host on port 28015:
+   mongosh --host mongodb0.example.com --port 28015
 
-     .. code-block:: sh
+.. note:: Connect to |service-fullname|
 
-        mongosh --host mongodb0.example.com --port 28015
+   If your remote host is a |service| cluster, you can copy your
```
mdb-ashley, line 84: [suggestion] Note that this renders as "a Atlas cluster".

source/connect.txt
```diff
@@ -38,133 +37,118 @@ This is equivalent to the following command:
...
+Specify different connection options to connect to different types of
+deployments.
 
-   To connect to a remote MongoDB instance and authenticate against 
-   the ``admin`` database as user ``alice``:
+Connect With Authentication
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-   .. code-block:: sh
+To connect to a MongoDB deployment requires authentication, use the
```
mdb-ashley, line 99: [typo] "...deployment that requires..."

source/connect.txt
```diff
@@ -38,133 +37,118 @@ This is equivalent to the following command:
...
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-To connect to a replica set:
+To connect to a replica set, perform one of the following actions:
 
-- If you are using the :manual:`DNS Seedlist Connection Format 
-  </reference/connection-string/#dns-seedlist-connection-format>`, you 
-  can include the ``+srv`` modifier in your connection string.
+- If you use the :manual:`DNS Seedlist Connection Format
+  </reference/connection-string/#dns-seedlist-connection-format>`,include
```
mdb-ashley, line 125: [typo] Missing a space after the comma.

source/connect.txt
```diff
@@ -38,133 +37,118 @@ This is equivalent to the following command:
...
+     :manual:`Enable Access Control </tutorial/enable-authentication/>`.
+
+   - To provision access to a MongoDB deployment, see :manual:`Database
+     Users </core/security-users/>`.
 
 Connect to a Replica Set
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-To connect to a replica set:
+To connect to a replica set, perform one of the following actions:
```
mdb-ashley, line 122: [suggestion] I'm finding this structure a little confusing. What are the two options here? Should each option have its own heading?

source/connect.txt
```diff
@@ -38,133 +37,118 @@ This is equivalent to the following command:
...
-     To connect to a three-member replica set named ``replA``:
-
-     .. code-block:: sh
+  .. code-block:: sh
 
-        mongosh "mongodb://mongodb0.example.com.local:27017,mongodb1.example.com.local:27017,mongodb2.example.com.local:27017/?replicaSet=replA"
+     mongosh "mongodb://mongodb0.example.com.local:27017,mongodb1.example.com.local:27017,mongodb2.example.com.local:27017/?replicaSet=replA"
 
-.. note::
+.. note:: directConnection Parameter
```
mdb-ashley, line 151: [questions] Does this note apply to only the second option above, or to both? It's not clear until you read the whole note. Also, should we title this note a little more specifically? Perhaps, "directConnection Parameter added automatically" or something like that?

source/connect.txt
```diff
@@ -21,12 +21,11 @@ Prerequisites
 .. include:: /includes/fact-connect-prereqs.rst
```
mdb-ashley, line 21: [question] Does this include need an update? It says 4.0 or greater, but I'm not sure if we might be indicating what's possible here, rather than what's supported.

source/connect.txt
```diff
@@ -38,133 +37,118 @@ This is equivalent to the following command:
...
 
-     .. code-block:: sh
+.. code-block:: sh
 
-        mongosh --port 28015
+   mongosh --port 28015
 
-MongoDB Instance on a Remote Host
----------------------------------
+MongoDB Deployment on a Remote Host
```
mdb-ashley, line 60: [suggestion] We removed "MongoDB" from other headings -- should we remove it here?

source/connect.txt
```diff
@@ -21,12 +21,11 @@ Prerequisites
 .. include:: /includes/fact-connect-prereqs.rst
    
 
-Local MongoDB Instance on Default Port
---------------------------------------
+Local Deployment on Default Port
```
mdb-ashley, line 24: [suggestion] Each of these headings is really a mini-task. Would it make sense to title all of them with verb phrases, like tasks? 

For example:
- "Connect to a Local Deployment"
- "Connect to a Remote Deployment"
- "Connect with Options"

source/connect.txt
```diff
@@ -267,17 +250,17 @@ To disconnect from a deployment and exit ``mongosh``, you can:
...
+  - ``authMechanismProperties=CANONICALIZE_HOST_NAME:none``
 
 - ``mongosh`` currently only supports the ``zlib``
   :manual:`compressor </core/wiredtiger/#compression>`. The following 
   compressors are not supported:
 
   - ``zstd``
   - ``snappy``
-
-
```
mdb-ashley, line 283: [suggestion] Looking at this page from an information typing perspective, I'm not sure all these tasks actually belong on this page. Something to perhaps file a ticket about and consider later.

mdb-ashley (changes requested):
Nice work, Jeff! This page looks much better. I left a few comments for your consideration. Let me know what you think.

jeff-allen-mongo:
@mdb-ashley thanks very much for the review. This is ready for another look whenever you have a chance

source/connect.txt
```diff
@@ -119,43 +120,55 @@ database as user ``alice``:
 Connect to a Replica Set
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-To connect to a replica set, perform one of the following actions:
+To connect to a replica set, you can either:
```
mdb-ashley, line 123: [praise] Nice change! I appreciate the minimalism and friendly tone.

mdb-ashley (approved):
Nice changes, Jeff! This one LGTM. Thanks for handling this.

source/connect.txt
```diff
@@ -38,204 +37,200 @@ This is equivalent to the following command:
...
 
-      mongosh "mongodb://mongodb0.example.com:28015" --username alice --authenticationDatabase admin
+For example, to authenticate on the ``admin`` database as user
+``alice``, run the following command:
 
-.. note::
+.. code-block:: sh
 
-   To provide a password with the command instead of using the masked 
-   prompt, you can use the ``--password`` option.
```
addaleax, line 125: We don’t want to tell users about `--password` anymore?

source/connect.txt
```diff
@@ -38,204 +37,200 @@ This is equivalent to the following command:
...
+
+   ``mongosh`` automatically adds the ``directConnection=true``
+   parameter to the connection string, regardless of the connection
+   string format used, unless at least one of the following is true:
 
    - The ``replicaSet`` query parameter is present in the connection string.
-   - The connection string uses the ``mongodb+srv://`` scheme.
+   - The connection string uses the ``mongodb+srv://`` connection string
+     format.
    - The connection string contains a seed list with multiple hosts.
```
addaleax, line 172: Maybe this is obvious, but:

```suggestion
   - The connection string already contains a `directConnection` parameter.
```

source/connect.txt
jeff-allen-mongo: I wasn't sure about this. I removed it because users are still prompted to specify the password when they omit the ``--password`` option, and not using the ``--password`` option ensures that the password is never viewable as plaintext on the user's screen. I can add this note back in if we still want users to be aware of this slightly less secure method of specifying the password?

addaleax: Yeah, I’ll leave this up to you. The idea isn’t really that human users, which I assume this guide here is targeting, would use it, and more that it’s a useful flag to provide for programmatic usage of mongosh.

jeff-allen-mongo: Ah, that's a good point about programmatic usage. I'll add the note back in, because it's likely good for drivers users to be aware of.

PR #252, opened 2023-03-23
remove unnecessary (and deprecated) class directives

schmalliso:
Toctrees no longer produce visible output so the `.. class:: hidden` directive isn't doing anything anyways. We're removing support for it in a forthcoming Snooty release, so might as well clean it up and avoid a warning in your logs in advance. Cheers!


jeff-allen-mongo:
@schmalliso sorry I missed the initial ping on this one! Retroactive LGTM 🚀

PR #317, opened 2024-02-09
DOCSP-36565 fixing mongosh installation .repo file

ltran-mdb2:
## DESCRIPTION

changing $releasever to 2023 for .repo file

## STAGING

https://preview-mongodbltranmdb2.gatsbyjs.io/mongodb-shell/DOCSP-36565/install/#procedure-3

## JIRA

https://jira.mongodb.org/browse/DOCSP-36565

## BUILD LOG

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=65c68f946b2f17c95aa79bec

## Self-Review Checklist

- [x] Is this free of any warnings or errors in the RST?
- [x] Is this free of spelling errors?
- [x] Is this free of grammatical errors?
- [x] Is this free of staging / rendering issues?
- [x] Are all the links working?

### DIFF
--- a/source/includes/steps-install-shell-linux-rpm.yaml
+++ b/source/includes/steps-install-shell-linux-rpm.yaml
@@ -24,7 +24,7 @@ content: |
 
                [mongodb-org-{+mdb-version+}]
                name=MongoDB Repository
-               baseurl=https://repo.mongodb.org/yum/redhat/2023/mongodb-org/{+mdb-version+}/$basearch/
+               baseurl=https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/{+mdb-version+}/$basearch/
                gpgcheck=1
                enabled=1
                gpgkey=https://www.mongodb.org/static/pgp/server-{+pgp-version+}.asc
--- a/source/includes/steps-install-shell-linux-rpm.yaml
+++ b/source/includes/steps-install-shell-linux-rpm.yaml
@@ -24,7 +24,7 @@ content: |
 
                [mongodb-org-{+mdb-version+}]
                name=MongoDB Repository
-               baseurl=https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/{+mdb-version+}/$basearch/
+               baseurl=https://repo.mongodb.org/yum/redhat/2023/mongodb-org/{+mdb-version+}/$basearch/
                gpgcheck=1
                enabled=1
                gpgkey=https://www.mongodb.org/static/pgp/server-{+pgp-version+}.asc
@@ -50,7 +50,7 @@ content: |
 
                [mongodb-org-{+mdb-version+}]
                name=MongoDB Repository
-               baseurl=https://repo.mongodb.org/yum/amazon/$releasever/mongodb-org/{+mdb-version+}/$basearch/
+               baseurl=https://repo.mongodb.org/yum/amazon/2023/mongodb-org/{+mdb-version+}/$basearch/
                gpgcheck=1
                enabled=1
                gpgkey=https://www.mongodb.org/static/pgp/server-{+pgp-version+}.asc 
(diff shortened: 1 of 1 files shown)


source/includes/steps-install-shell-linux-rpm.yaml
```diff
@@ -50,7 +50,7 @@ content: |
 
                [mongodb-org-{+mdb-version+}]
                name=MongoDB Repository
-               baseurl=https://repo.mongodb.org/yum/amazon/$releasever/mongodb-org/{+mdb-version+}/$basearch/
+               baseurl=https://repo.mongodb.org/yum/amazon/2023/mongodb-org/{+mdb-version+}/$basearch/
                gpgcheck=1
                enabled=1
                gpgkey=https://www.mongodb.org/static/pgp/server-{+pgp-version+}.asc
```
ajhuh-mdb, line 56: [suggestion] I can't comment directly on the line, but on line 63 it might help to add `2023` as an example for Amazon Linux version since that's what's used in the example above:

> Amazon Linux version (for example, ``2023``)

ajhuh-mdb (approved):
LGTM % one suggestion!

source/includes/steps-install-shell-linux-rpm.yaml
ltran-mdb2: Done!

source/includes/steps-install-shell-linux-rpm.yaml
```diff
@@ -50,7 +50,7 @@ content: |
 
                [mongodb-org-{+mdb-version+}]
                name=MongoDB Repository
-               baseurl=https://repo.mongodb.org/yum/amazon/$releasever/mongodb-org/{+mdb-version+}/$basearch/
+               baseurl=https://repo.mongodb.org/yum/amazon/2023/mongodb-org/{+mdb-version+}/$basearch/
```
addaleax, line 53: It seems like right now this is a template for all AL versions, we should make clear that the user needs to adjust this value depending on the actual version on their system

source/includes/steps-install-shell-linux-rpm.yaml
```diff
@@ -24,7 +24,7 @@ content: |
 
                [mongodb-org-{+mdb-version+}]
                name=MongoDB Repository
-               baseurl=https://repo.mongodb.org/yum/redhat/$releasever/mongodb-org/{+mdb-version+}/$basearch/
+               baseurl=https://repo.mongodb.org/yum/redhat/2023/mongodb-org/{+mdb-version+}/$basearch/
```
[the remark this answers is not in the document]
ltran-mdb2: thanks for catching this! I switched back to $releasever and the installation worked as expected

source/includes/steps-install-shell-linux-rpm.yaml
ltran-mdb2: My thought is that the manual instructions underneath (starting with "You can also download...") indicate that the users can switch depending on their version. Do you think it should be emphasized more?

addaleax: Yeah, I think so, the manual downloads don't really cover the same use case, and we already know that users don't have a particularly easy time figuring out how to pick the "right" way to install the MongoDB server through apt/yum.

addaleax: For what it's worth, you don't need to consider this a blocking issue, I'm just worried that this will lead to users accidentally using the 2023 repo when they didn't intend to do that, maybe that's just fine and maybe AL2023 is already widely adopted enough that it doesn't really matter

ltran-mdb2: I just added a note under the .repo file about Amazon Linux 2 - what do you think?

addaleax (approved):
LGTM!

PR #335, opened 2024-06-21
DOCSP-40374: 2.2.9 changelog updates

sarah-olson-mongodb:
## DESCRIPTION
2.2.9 changelog updates. 

Note to reviewers: This release does not include [user facing changes](https://mongodb.slack.com/archives/CUHC9R8J0/p1718378038131509). I've mentioned the node.js driver version bump, and included a link to the full release notes on consistency grounds. We generally include the release notes link, and I think it may be confusing to omit. Happy to delete if you disagree.

## STAGING
https://preview-mongodbsaraholsonmongodb.gatsbyjs.io/mongodb-shell/DOCSP-40374/changelog/#v2.2.9

## JIRA
https://jira.mongodb.org/browse/DOCSP-40374

## BUILD LOG
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=6675e75bebce19f314f71a30

## Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?
- [ ] Is this free of spelling errors?
- [ ] Is this free of grammatical errors?
- [ ] Is this free of staging / rendering issues?
- [ ] Are all the links working?

### DIFF
--- a/source/changelog.txt
+++ b/source/changelog.txt
@@ -17,7 +17,7 @@ v2.2.9
 
 *Released June 14, 2024*
 
-``mongosh`` now uses version 6.5.0 of the driver:`Node.js driver </node/>`.
+``mongosh`` now uses version 6.5.0 of the :driver:`Node.js driver </node/>`.
 
 `Full release notes available on JIRA 
 <https://jira.mongodb.org/secure/ReleaseNote.jspa?projectId=17381&version=39810>`__.
--- a/source/changelog.txt
+++ b/source/changelog.txt
@@ -17,7 +17,7 @@ v2.2.9
 
 *Released June 14, 2024*
 
-``mongosh`` now uses version 6.5.0 of the :driver:`Node.js driver </node/>`.
+``mongosh`` now uses version 6.7.0 of the :driver:`Node.js driver </node/>`.
 
 `Full release notes available on JIRA 
 <https://jira.mongodb.org/secure/ReleaseNote.jspa?projectId=17381&version=39810>`__.
--- a/source/changelog.txt
+++ b/source/changelog.txt
@@ -12,6 +12,16 @@ Release Notes
    :depth: 1
    :class: singlecol
 
+v2.2.9
+------
+
+*Released June 14, 2024*
+
+``mongosh`` now uses version 6.5.0 of the driver:`Node.js driver </node/>`.
+
+`Full release notes available on JIRA 
+<https://jira.mongodb.org/secure/ReleaseNote.jspa?projectId=17381&version=39810>`__.
+
 v2.2.5
 ------
 
--- a/source/changelog.txt
+++ b/source/changelog.txt
@@ -19,9 +19,6 @@ v2.2.9
 
 ``mongosh`` now uses version 6.7.0 of the :driver:`Node.js driver </node/>`.
 
-`Full release notes available on JIRA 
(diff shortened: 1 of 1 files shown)


source/changelog.txt
```diff
@@ -12,6 +12,16 @@ Release Notes
    :depth: 1
    :class: singlecol
 
+v2.2.9
+------
+
+*Released June 14, 2024*
+
+``mongosh`` now uses version 6.5.0 of the :driver:`Node.js driver </node/>`.
```
ianf-mongodb, line 20: Should this be 6.7.0? There is already a RN from v2.20 that states mongosh is using 6.5.0: 
https://preview-mongodbsaraholsonmongodb.gatsbyjs.io/mongodb-shell/DOCSP-40374/changelog/#v2.2.0.

I also dont see any notice of the driver bump and the latest release page: https://github.com/mongodb-js/mongosh/releases/tag/v2.2.9

```suggestion
``mongosh`` now uses version 6.7.0 of the :driver:`Node.js driver </node/>`.
```

ianf-mongodb (commented):
Hi @sarah-olson-mongodb one question regarding the driver version.

source/changelog.txt
addaleax: Yes :)

sarah-olson-mongodb: Thanks a ton @ianf-mongodb! Updated!

sarah-olson-mongodb: I think it makes sense to omit the Jira release notes then. As Anna mentioned, they don't contain customer facing changes, and as Ian mentioned, they don't contain the node.js driver update either.

addaleax (approved):

source/changelog.txt
ianf-mongodb: Just to be clear, what I meant was if we publish these RN's, the Github release page should match the driver bullet. For example for version 1.8 both [the RN's ](https://github.com/mongodb-js/mongosh/releases/tag/v1.8.0)and the [release page](https://github.com/mongodb-js/mongosh/releases/tag/v1.8.0) mention the driver bump. Would it be possible to add this to the release page @addaleax?

addaleax: @ianf-mongodb So ... in an ideal world we'd have this, but at least as long as we don't manually curate release notes for Github, this isn't going to show up automatically. In this case, we've had two failed releases (hence skipping versions 2.2.7 and 2.2.8), and for the change that bumped the driver version, the engineer creating it marked it as a "no direct user impact" change (which is, for the most part, accurate – it's a dependency bump, not an API change).

We can do a one-off and edit the GH release notes to be more expressive here, but for mongosh, we haven't really done this; if you think that it's helpful for us to do this, then we should probably integrate this into our release process as well.

addaleax: So – for a bit more context, I already mentioned this to Sarah on Slack, but TSEs have historically advocated specifically for logging Node.js driver version updates in changelogs because that way, they would be able to identify which driver version mongosh is using when a user reports a bug (and only specifies the mongosh version, which is reasonable for a mongosh user) and then potentially try to figure out whether the issue originated from the driver or not. That's why this PR was created in this form in the first place, rather than just saying "no impact to users", even though it _should_ have no impact to users.

ianf-mongodb: LGTM pending we update the process for mongosh release notes to reflect this workflow on: https://wiki.corp.mongodb.com/display/DE/mongosh+Release+Notes

ianf-mongodb (approved):

PR #364, opened 2024-11-01
DOCSP-45058-mongosh-undefined-behavior

amalhotra-mdb:
## DESCRIPTION
[DOCSP-44560](https://jira.mongodb.org/browse/DOCSP-44560) notes that `mongosh` converts undefined to null during write operations, unlike the legacy `mongo` shell; this is not properly documented in the mongosh docs, so this PR adds this info to the [Compatibility Changes w/ Legacy `mongo` shell page](https://www.mongodb.com/docs/mongodb-shell/reference/compatibility/#compatibility-changes-with-legacy-mongo-shell)

## STAGING
https://deploy-preview-364--docs-mongodb-shell.netlify.app/reference/compatibility/#undefined-values

## JIRA
https://jira.mongodb.org/browse/DOCSP-45058

## Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?
- [ ] Is this free of spelling errors?
- [ ] Is this free of grammatical errors?
- [ ] Is this free of staging / rendering issues?
- [ ] Are all the links working?

### DIFF
--- a/source/reference/compatibility.txt
+++ b/source/reference/compatibility.txt
@@ -202,6 +202,37 @@ in ``mongosh`` better align with the types used by the MongoDB Drivers.
    For more information on managing types, refer to the
    :manual:`schema validation overview </core/schema-validation>`.
 
+Undefined Values
+----------------
+
+The undefined BSON type is deprecated <https://bsonspec.org/spec.html>__ in
+:binary:`bin.mongosh`. If you insert a document with the undefined BSON type in
+:binary:`bin.mongosh`, your deployment stores a null value. To create `undefined` 
+fields in your document, use the legacy :binary:`~bin.mongo` shell.
+
+For example, consider running the same code to insert a document in both
+:binary:`bin.mongosh` and in the legacy :binary:`~bin.mongo` shell:
+
+.. code-block:: javascript
+
+   db.people.insertOne( { name : "Sally", age : undefined } )
+
+When you run this code to insert a document with the undefined BSON type in
+:binary:`bin.mongosh`, the shell inserts the following document: 
+
+.. code-block:: javascript
+   "copyable: false
+
+   { name : "Sally", age : null }
+
+However, when you run the same code to insert a document with the undefined BSON type in
+the legacy :binary:`~bin.mongo` shell, the shell inserts the following document: 
+
+.. code-block:: javascript
+   "copyable: false
+
+   { name : "Sally", age : undefined }
+
 Object Quoting Behavior
 -----------------------
 
--- a/source/reference/compatibility.txt
+++ b/source/reference/compatibility.txt
@@ -208,7 +208,8 @@ Undefined Values
 The undefined BSON type is `deprecated <https://bsonspec.org/spec.html>`__ in
 :binary:`bin.mongosh`. If you insert a document with the undefined BSON type in
 :binary:`bin.mongosh`, your deployment replaces the undefined value in your 
-document with the BSON null value.
+document with the BSON null value. There is no supported way of inserting undefined
+values into your database using :binary:`bin.mongosh`.
 
 For example, consider running the same code to insert a document in both
 :binary:`bin.mongosh` and in the legacy :binary:`~bin.mongo` shell:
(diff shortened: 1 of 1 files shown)


source/reference/compatibility.txt
```diff
@@ -202,6 +202,37 @@ in ``mongosh`` better align with the types used by the MongoDB Drivers.
    For more information on managing types, refer to the
    :manual:`schema validation overview </core/schema-validation>`.
 
+Undefined Values
+----------------
+
+The undefined BSON type is `deprecated <https://bsonspec.org/spec.html>`__ in
+:binary:`bin.mongosh`. If you insert a document with the undefined BSON type in
+:binary:`bin.mongosh`, your deployment stores a null value. To create undefined
```
sarah-olson-mongodb, line 210: (question)
Do you think we could replace "stores a null value" with something a little less jargony, that retains the same meaning? Maybe add something about the impact to to document or to data when undefined is converted to null?

source/reference/compatibility.txt
```diff
@@ -202,6 +202,37 @@ in ``mongosh`` better align with the types used by the MongoDB Drivers.
...
+fields in your documents, use the legacy :binary:`~bin.mongo` shell.
```
sarah-olson-mongodb, line 211: (suggestion)
Maybe omit this last sentence because I don't think we want to point folks to the legacy shell. Is there maybe another solution?

source/reference/compatibility.txt
```diff
@@ -202,6 +202,37 @@ in ``mongosh`` better align with the types used by the MongoDB Drivers.
...
+
+   { name : "Sally", age : null }
+
+However, when you run the same code to insert a document with the undefined BSON type in
+the legacy :binary:`~bin.mongo` shell, the shell inserts the following document: 
+
+.. code-block:: javascript
+   "copyable: false
+
+   { name : "Sally", age : undefined }
```
sarah-olson-mongodb, line 234: (question) 
I'm curious about whether there's another workaround. It might be good to check with @addaleax during tech review.

sarah-olson-mongodb (commented):
Thanks so much for this update @amalhotra-mdb! And fantastic job identifying this doc gap and creating a ticket to resolve it! This all looks really good. 

My main question is around whether we want to point folks to the legacy mongo shell as the workaround? I think you might want to check in with @addaleax or @GaurabAryal, if you haven't already, regarding whether there might be another workaround. 

I'll be OOO next week and you don't need to wait on me to get this internal review LGTM'd. :)

source/reference/compatibility.txt
amalhotra-mdb: After a bit of investigation, it seems that generally users should never actually need to insert an undefined value, so I'm thinking we might be okay to not provide a work around? I will get more opinions on this though. It seems if we really want to point them at a workaround, they could use a driver that supports inserting raw BSON data, e.g. through a Python script

nvillahermosa-mdb (approved):
Short and clear, LGTM

source/reference/compatibility.txt
```diff
@@ -202,6 +202,30 @@ in ``mongosh`` better align with the types used by the MongoDB Drivers.
...
    :manual:`schema validation overview </core/schema-validation>`.
 
+Undefined Values
+----------------
+
+The undefined BSON type is `deprecated <https://bsonspec.org/spec.html>`__ in
+:binary:`bin.mongosh`. If you insert a document with the undefined JS type in
+:binary:`bin.mongosh`, your deployment replaces the undefined value in your 
+document with the BSON null value. There is no supported way of inserting undefined
+values into your database using :binary:`bin.mongosh`.
```
FGasper, line 212: It‘s also relevant that mongosh will represent any stored BSON undefined values as null. So if you’ve stored a BSON undefined value with some other tool—e.g., Go’s driver or the `mongo` shell—then you fetch that document with mongosh, you’ll see null where the actual stored document has undefined.

Should this section mention that as well?

source/reference/compatibility.txt
```diff
@@ -202,6 +202,30 @@ in ``mongosh`` better align with the types used by the MongoDB Drivers.
    For more information on managing types, refer to the
    :manual:`schema validation overview </core/schema-validation>`.
 
+Undefined Values
+----------------
+
+The undefined BSON type is `deprecated <https://bsonspec.org/spec.html>`__ in
+:binary:`bin.mongosh`. If you insert a document with the undefined JS type in
```
FGasper, line 209: BSON undefined is deprecated everywhere, not just in mongosh.

